PR #4, opened 2015-10-15
Added Form Editor 

pefernan:


csadilek:
Something is wrong here ;). The PR carries 101 commits.

PR #6, opened 2015-12-18
Fixing Issues related ListBoxes

pefernan:
- Fixed issue setting default value to ListBoxes
- Added Text property on ListBox options
- Refactored source generation to let a Field define more fields on the FormView
- Small cleanup on FormView API


csadilek:
Thanks!

master: 9a9a91815cf7275c871dc259272e76a5b88d0421

PR #15, opened 2016-04-19
Fix for Model2FormTransfomerServiceImplTest

mbarkley:


csadilek:
+1

manstis:
IDK if it's of any interest to LiveSpark but kie-wb-commons now has an "Import Example" module that uses a properties file and Wizard to expose import of arbitrary repositories into a Workbench. We've (well, me) replaced the "auto-clone" of example(s) repositories with this in the KIE Workbench distributions.

mbarkley:
@manstis Thanks for the tip. That will probably be of use to us in the future.

PR #16, opened 2016-04-22
Fix issues with boolean column generation and issue with build service cleanup.

mbarkley:
This PR depends on two PRs in [uberfire-extensions](https://github.com/uberfire/uberfire-extensions/pull/274) and [kie-wb-common](https://github.com/droolsjbpm/kie-wb-common/pull/348).

This PR:
* Uses the "is" prefix for getters of boolean values.
* Uses the new fully qualified name of CheckboxCellImpl (moved in linked PRs).
* Fixes exception when session expires and target directory is deleted before the deployed war file.
* Adds errai-ui as a dependency (no longer brought in transitively).


csadilek:
+1

PR #18, opened 2016-04-26
[ DO NOT MERGE ] Testing PR Builds for this repository

mbarkley:


psiroky:
@mbarkley I can reproduce the same compilation error locally, so this is likely not caused by the Jenkins environment.

psiroky:
So the build finished, but there are two test failures. Not sure what is the root cause of those.

mbarkley:
@psiroky I think the cause of this issue is that the build is done with `verify` rather than `install`. The Arquillian tests that run last test the war file installed by `livespark-webapp`, using maven to resolve the artifact.

Can we change this and try again?

psiroky:
We can change that (with few other things as well, as we don't want to install artifacts from PRs into default ~/.m2 repo).

mbarkley:
@psiroky FYI I've looked into these most recent failures. They are related to the -Dmaven.local.repo property you have configured for the build, but I think I can fix the test harness to work with a non-default local repo.

psiroky:
@mbarkley that would be great as the PR builds should use non-default local repo...

PR #19, opened 2016-04-29
Fixes to integration-test harness + fix a double-deployment bug

mbarkley:
The generated app will now be built using the local maven repository set by
the system property "maven.repo.local".

The integration tests are now configured to build the generated app
with the same maven local repo as the integration tests use themselves.

Avoid calling buildAndDeploy multiple times by only dispatching to
the super class a single time. This caused integration tests to intermittently fail
from the generated war file being deployed multiple times.


mbarkley:
jenkins, retest this

csadilek:
+1

PR #20, opened 2016-05-04
Moved from UiBinder to Errai UI

pefernan:
- Moved from UiBinder components to Errai UI
- Removed unused constants
- Fixed ClassCastException
- Removed _ on translation keys


livespark-common-widgets/livespark-crud-component/src/main/java/org/livespark/widgets/crud/client/resources/i18n/CrudComponentConstants.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package org.livespark.widgets.crud.client.resources.i18n;
+
+import org.jboss.errai.ui.shared.api.annotations.TranslationKey;
+
+public interface CrudComponentConstants {
+
+    @TranslationKey( defaultValue = "")
+    String CrudComponentViewImpl_newInstanceButton = "CrudComponentViewImpl.newInstanceButton";
```
mbarkley: Using underscores in identifiers is deprecated in Java 8 (and will show warnings in the build). I see why you did it in this case, but I think it's best to remove them.

mbarkley:
Hi @pefernan 

As I mentioned in-line, I think we should get rid of the underscores in the field names of the translation key constants since they will cause build warnings.

Otherwise looks good.

pefernan:
@mbarkley tbh I thought it was only with variables called "_" but you're right... so removed them.

mbarkley:
Jenkins, retest this.

mbarkley:
+1

PR #24, opened 2016-05-20
Add producer for *.client.local.* and generated class filters.

mbarkley:
This PR filters out generated *.java files from the DataModelerService and prevents the KieBuilder from compiling *.client.local.* classes.


csadilek:
+1

PR #28, opened 2016-06-07
Enums support & providing service

pefernan:
- Added field providing mechanism
- Added support first version I (not enabled for static forms)

- Fixed issues field providing mechanism
- Added support first version II

- Moved modules
- Added some testing


manstis:
Pesonally, -1 this PR (nothing wrong with the code though)

We have ```DataModelOracle``` in ```kie-wb-common``` that has enum support (and will be used for dynamic bean validation in Form Modeller - in the near future). I'd like to stop the temptation to write another "type oracle" implementation early as it'll be better to fix what's broken (with the current DMO to be used for Forms) than to retro fit Form Modeller with the existing DMO in the future (as I don't think it'd ever happen and we'd have two "type metadata" services).

.. but this is my opinion (I think everyone knows this already!!).

pefernan:
@manstis I agree wit using DMO, but right now I do require inspecting the models that are part of the application codebase, not only the ones created within the workbench. I think the currenyt DMO only supports the second scenario. Do you think is it possible to have that feature too? I'll be happy to migrate all the model processing to DMO as soon as I could do that.

pefernan:
@manstis do you think that we could have a call tomorrow to talk about that?

mbarkley:
@pefernan FYI, I tried building from master (i.e. without this PR) and got the same build error as Jenkins, so that is unrelated to your change. I will look into this more.

pefernan:
@mbarkley Ok, I'll take a look tomorrow morning to see what's wrong. Let me know if you find something

mbarkley:
@pefernan The test errors were caused by a type in the default package in users-new. I've sent a PR to your livespark-playground fork.

pefernan:
@mbarkley ops! I was aware of that class on the default package, I had planned to fix it this morning, so you did the job for me ;-) thanks!

livespark-form-modeler/livespark-form-modeler-api/pom.xml
```diff
@@ -57,6 +57,12 @@
       <groupId>org.uberfire</groupId>
       <artifactId>uberfire-layout-editor-api</artifactId>
     </dependency>
+
+    <!-- test -->
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+    </dependency>
```
psiroky: It seems this should be in a test scope?

pefernan: @psiroky you're right, I'm fixing it, thanks!

manstis:
+1 to merge.

pefernan:
Jenkins, retest this

pefernan:
To clarify Mike's +1: We had a call and I'll be moving the annotation processing engine to DMO during the next days. I will require some refactor on DMO btw.

pefernan:
Jenkins, retest this

PR #29, opened 2016-06-10
- Support for text areas on dynamic forms

pefernan:


livespark-form-modeler/livespark-form-modeler-dynamic-renderer/livespark-form-modeler-dynamic-renderer-backend/src/main/java/org/livespark/formmodeler/renderer/backend/service/impl/processors/TextAreaAnnotationProcessor.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    protected TextAreaFieldDefinition buildFieldDefinition( Annotation annotation, FieldSetting setting ) {
+        if ( !supportsAnnotation( annotation )) {
+            return null;
+        }
+
+        TextArea textAreaAnnotation = (TextArea) annotation;
+        TextAreaFieldDefinition textArea = (TextAreaFieldDefinition) fieldManager.getFieldFromProvider( TextAreaFieldDefinition.CODE, setting.getTypeInfo() );
```
mbarkley: I'm not a fan of the casting here. I think it would be better to parameterize this method so that the cast is unnecessary. Looking at some of the other processors, it seems as though there is a one-to-one relationship between typeCodes and types extending FieldDefinition. If that is the case, can we use types instead of type codes and change the signature of the method? Something like this would be great:

    <F extends FieldDefinition> F getFieldFromProvider( Class<F> type, FieldTypeInfo typeInfo );

mbarkley:
+1

I've commented inline about some API I dislike, but it wasn't introduced in this commit and would require unrelated changes to fix so it isn't cause to delay merging.

PR #31, opened 2016-06-21
Fixes after IP BOM 7.0.0.CR4 upgrade

psiroky:
The build will fail because the changes are not yet on master -- and livespark PR builds are not configured to run upstream repos first.


csadilek:
+1

PR #40, opened 2016-07-19
JBPM-5071  - remove of jbpm runtime from workbench

pefernan:
Added missing changes to LivSpark WevApp from the JBPM-5071.


psiroky:
The livespark tests are very often (randomly?) failing on Jenkins. This makes the PR builds much less useful as basically all of them fail, and one needs to wonder if the fails are related to his changes or not. It would be great if someone could take a look at those tests and fix the stability issues.

csadilek:
+1

csadilek:
@psiroky yep, we'll look into it.

PR #41, opened 2016-07-20
Test fixes

mbarkley:
Fixes a broken test in LiveSparkEntryPointTest and moves the BuildAndDeploymentTests to a maven profile so that they do not run on Jenkins, but can be run locally.


csadilek:
+1

PR #43, opened 2016-07-27
Moved annotation processing mechanism to use DMO instead of reflection II. Fixing integration issues.

pefernan:
This fixes some CDI issues when injecting FormRenderingContextGenerator.


csadilek:
+1

PR #46, opened 2016-08-05
Upgrade to GWT 2.8.0-rc1

mbarkley:


livespark-showcase/livespark-webapp/pom.xml
```diff
@@ -43,6 +43,12 @@
   </properties>
 
   <dependencies>
+    <!-- drools-compiler has dependency on org.eclipse.jdt.core.compiler:ecj:jar:3.5.1:compile, see http://code.google.com/p/google-web-toolkit/issues/detail?id=4479 -->
```
psiroky: Is this comment still relevant? drools-compiler now depends on ECJ 4.4.3 and I am not sure the original applies anymore.

manstis: I was thinking much the same thing over at https://github.com/droolsjbpm/drools-wb/pull/263/files#diff-7a8bb56e50632e02768371ed4ddc1688R42

csadilek: @psiroky @manstis can't comment on ecj for drools but this is still needed so the right version of jdt is on the classpath during gwt compilation. Would you prefer I remove the comment on all PRs?

psiroky: +1 to remove the comment as it only confuses things (IMO).

mbarkley:
Removed the ecj comment.

csadilek:
+1

PR #47, opened 2016-08-11
Remove livespark-integration-tests.

mbarkley:
The LS integration tests are flaky and depend on the contents of [this repo](https://github.com/pefernan/livespark-playground) being in sync with the LS version being run.

I think it is best to remove them for now and replace them with other integration tests at a later date when the project is more stable.

@csadilek What do you think?


pefernan:
+1

csadilek:
I think it makes sense as the main point of these integration tests is to test the build process which will be replaced by the new provisioning service.

PR #51, opened 2016-09-12
Validate model that is bound to UI, not paused model that is out of sync.

mbarkley:
@pefernan Can you please review this to verify that it fixes the problem we discussed?


mbarkley:
@pefernan I've pushed another commit that fixes the NumberFormatException you showed me.

The issue was that the default converters in LS did not have special logic for handling the empty String. I've removed the converters that convert to or from String because Errai provides converters for those types that handle the empty String properly.

pefernan:
+1 with this change works fine. Thanks!

PR #52, opened 2016-09-13
Remove CDI AppFlow API and use AsyncDataProvider as ListView input

mbarkley:
This PR has two changes:

1. Removing the CDI API for AppFlow.
2. Changing the ListView to use an AsyncDataProvider<T> as input instead of a List<T>.

The reason for (1) is that the implementation for this CDI API has lots of limitations. It would require lots of work to smooth out all the unintuitive wrinkles and I don't believe it would be worth the trouble. Instead ListView and other UI components used in AppFlows can implement an interface (`UIComponent`).

(2) is a necessary enhancement, so that a ListView can be displayed by the AppFlow engine before the data it displays is loaded.


csadilek:
+1

PR #54, opened 2016-10-04
Add back mistakenly deleted method in BaseEntityService.

mbarkley:
Fixes a compilation issue in the generated LS app.


csadilek:
+1

PR #63, opened 2016-11-30
AppFormer rename

mbarkley:
This PR is the first step towards creating an app-former repository. In this commit I have:

* Renamed all group IDs, Java package names, and GWT modules of the from `org.livespark.*` to `org.kie.appformer.*`
* Renamed all maven artifact IDs from `livespark-*` to `app-former-*`
* Renamed classes with `LiveSpark` in their names (replaced with `AppFormer`)
* Removed any other textual reference (comments, HTML markup, etc.) that reference LiveSpark
* Updated the example playground repository from livespark-playground to an updated version called [app-former-playground](https://github.com/mbarkley/app-former-playground)
* Fixed some errors that previously existed on master (I would have sent these in a separate PR, but I discovered them after having completed lots of renaming)

Regardless of whether or not this repository becomes the final home of app-former, this change set is necessary so that we can manually deploy app-former artifacts (the lack of these artifacts is blocking @Salaboy's docker/openshift integration efforts).

/cc @manstis @porcelli @pefernan @ederign @wmedvede


app-former-deployment/src/test/java/org/kie/appformer/deployment/tests/RestPipelineImplTest.java
```diff
@@ -272,7 +274,7 @@ public Properties getProperties() {
                 put("create-repo", "true");
                 put("branch", "master");
                 put("out-dir", tempPath.getAbsolutePath());
-                put("origin", "https://github.com/mbarkley/livespark-playground");
+                put("origin", "https://github.com/mbarkley/app-former-playground");
```
manstis: @ge0ffrey Further to our discussions this afternoon, there are a myriad of "example" repositories.. beware the dragons!

app-former-showcase/app-former-webapp/src/main/resources/example-repositories.properties
```diff
@@ -13,4 +13,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-ls-playground=https://github.com/mbarkley/livespark-playground.git
+ls-playground=https://github.com/mbarkley/app-former-playground.git
```
manstis: @ge0ffrey So you'd need to make sure any changes you make do not impact others' projects (e.g. you might need a "tag" for "experimental" or something too).

app-former-deployment/src/test/java/org/kie/appformer/deployment/tests/RestPipelineImplTest.java
ge0ffrey: Thanks Michael :) @mcimbora see above.

I have a solution for everything, I think. I 'll explain in a jira.

PR #64, opened 2016-12-06
Don't deploy client-side depenencies in app-former-webapp.

mbarkley:


mbarkley:
@Salaboy This remove lots of client-side jars from the app-former-webapp war (including the datasource management jars that caused your stracktraces). Could you please review this?

/cc @manstis

app-former-showcase/app-former-webapp/pom.xml
```diff
@@ -50,6 +50,18 @@
     </dependency>
 
     <dependency>
+      <groupId>org.uberfire</groupId>
+      <artifactId>uberfire-widgets-core-client</artifactId>
+      <scope>provided</scope>
```
manstis: @jsoltes They are only needed at compilation time of the webapp; when GWT needs the "client-side" sources to generate the JavaScript. They are customarily set to ```provided``` scope so they are not bundled in the WAR.

PR #70, opened 2017-01-27
Bug fixing

wmedvede:
- fixes a problem when there 2 or more entities that breaks the RPCProxies initialization
    - fixes the local deployment when an external datasource is selected


wmedvede:
Jenkins please retest this.

PR #77, opened 2017-03-20
GUVNOR-2933: Refactor to use the right APIs

pefernan:
- Adding some control for DatePicker on static forms.

@wmedvede as talked on irc... can you please take a look?


wmedvede:
Hello @pefernan  +1
Thanks!

PR #80, opened 2017-04-10
First iteration of Stunner integration with the AppFormer demo.

mbarkley:
Uses stunner diagrams as the primary mechanism of creating
UI flows for runtime execution.

Supports limited conditional behaviour and multistep forms.

Demo works with playground at this commit:
https://github.com/mbarkley/experimental-appformer-playground/commit/a4cbf4a8155a2c3d243bc3708ae91e3a899d4532


mbarkley:
@porcelli @ederign @pefernan @wmedvede 

Figured I would send a PR for this before the diff gets even bigger. I'd appreciate any reviews you guys have time for (but I understand it's a lot to take in!).

The `appformer-stunner-flowset-interpreter` module has an `Interpreter` class that converts a stunner diagram to an executable AppFlow. It has some tests to give you an idea of how it works.

Most of the other code introduced is stunner shape definitions copied and modified from the BPMN set. It still needs some cleaning up, but I think it's better to merge what is working now and do a more thorough clean up later once we have everything working.

mbarkley:
Closing this because it turns out AppFormer was still on 7.0.0-SNAPSHOT. I will fix that issue and send a PR with the stunner integration later.

PR #90, opened 2017-06-22
Current appformer.net  (designed by Entando)

ederign:
@pefernan is there any reason for having this Java project here?


mbarkley:
@ederign Why is this being committed here? Shouldn't this go in its own repo as a github site?

ederign:
@mbarkley agree

PR #92, opened 2017-07-04
RHBPMS-4805: Start form won't render, if it contains fields of BigDecimal, Character, Byte or Short type.

pefernan:
RHBPMS-4805: Start form won't render, if it contains fields of BigDecimal, Character, Byte or Short type.

This PR is part of an ensemble:
https://github.com/kiegroup/kie-wb-common/pull/960
https://github.com/kiegroup/appformer/pull/92


pefernan:
@jsoltes  @mbarkley could you take a look?

appformer-form-modeler/appformer-form-modeler-codegen/appformer-form-modeler-codegen-impl/src/main/java/org/kie/appformer/formmodeler/codegen/view/impl/java/inputs/util/SourceGenerationValueConvertersFactory.java
```diff
@@ -0,0 +1,56 @@
...
+package org.kie.appformer.formmodeler.codegen.view.impl.java.inputs.util;
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.kie.workbench.common.forms.model.FieldDefinition;
+
+public class SourceGenerationValueConvertersFactory {
```
mbarkley: Could you please add a javadoc or comment here that explains why we can lookup a converter with only one class name? I think intuitively one would expect that you lookup a converter with two types (the domain and target types). So it would be good to document why each type you lookup only has a single other type it converts to.

appformer-form-modeler/appformer-form-modeler-codegen/appformer-form-modeler-codegen-impl/src/test/java/org/kie/appformer/formmodeler/codegen/view/impl/inputs/impl/TextBoxHelperTest.java
```diff
@@ -0,0 +1,98 @@
...
+        testGetConverterClassName(String.class);
+    }
+
+    @Test
+    public void testGetConverterClassNameBigCharacter() {
+        testGetConverterClassName(Character.class);
+    }
+
+    @Test
+    public void testGetConverterClassNameBigChar() {
```
mbarkley: Based on the name of the JIRA, it seems like we should have regression tests here for Short, BigDecimal, Byte, etc.

pefernan:
@mbaluch better now? Added a test to verify that is using the right converter className instead of the old ones.

manstis:
Jenkins please retest this.

PR #107, opened 2017-11-30
RHBPMS-4926: Added Moment builder and fromNow formatting

nmirasch:
Modified the Moment interface to allow initialize with a date formatted and also make available the relative time fromNow formatting.

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/Moment.java
+++ b/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/Moment.java
@@ -49,6 +49,8 @@ public interface Moment {
 
     String format(String format);
 
+    String fromNow();
+
     String format();
 
     String toString();
@@ -73,5 +75,12 @@ public interface Moment {
 
         @JsMethod(namespace = JsPackage.GLOBAL)
         public static native Moment moment();
+
+        public static Moment moment(Long time) {
+            return moment(new Double(time));
+        }
+
+        @JsMethod(namespace = JsPackage.GLOBAL)
+        protected static native Moment moment(Double time);
     }
 }


uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/Moment.java
```diff
@@ -73,5 +75,8 @@ default Date asDate(){
 
         @JsMethod(namespace = JsPackage.GLOBAL)
         public static native Moment moment();
+
+        @JsMethod(namespace = JsPackage.GLOBAL)
+        public static native Moment moment(String dateStr, String pattern);
```
cristianonicolai: @nmirasch would it also work if you use these different constructors?
- https://momentjs.com/docs/#/parsing/unix-timestamp-milliseconds/
- https://momentjs.com/docs/#/parsing/date/
I guess these ones do not force you to format the date....

nmirasch:
@cristianonicolai changed the signature of minutes and fromNow function exposing. Removed the added constructor letting the moment construction with the relative minutes from the current moment. 
Could you take a look? Thanks!

cristianonicolai:
@nmirasch I think using the constructor is actually the best way to handle it, you could use any of those that I mentioned earlier. I might be wrong, but I think the milliseconds method in the Moment object only adjusts the millisecond part of the time, see: https://momentjs.com/docs/#/get-set/millisecond/

nmirasch:
@cristianonicolai Absolutely right, Changed to go back to the origin moment builder with the date format.

ederign:
@cristianonicolai can you please take a look on this?

cristianonicolai:
@ederign on it :)

@nmirasch ok so I remember the issue I had in the past with this API, in order to use the proper constructor method in JSNI, you need to use Double and not Long as the parameter. Check my example here, if that works for you: https://github.com/cristianonicolai/appformer/commit/6419b1a416836e15433d41c7810f8feebfeb2a46

nmirasch:
@cristianonicolai great! the double was the point. Thanks a lot
Added the change to the PR

PR #117, opened 2017-12-12
AF-934 - Asset upload / download to not work inside projects which na…

tsurdilo:
…me contains spaces

### DIFF
--- a/uberfire-server/src/main/java/org/uberfire/server/FileDownloadServlet.java
+++ b/uberfire-server/src/main/java/org/uberfire/server/FileDownloadServlet.java
@@ -49,7 +49,7 @@ public class FileDownloadServlet
         try {
 
             //See https://bugzilla.redhat.com/show_bug.cgi?id=1202926
-            final String encodedPath = FileServletUtil.encodeFileNamePart(request.getParameter("path"));
+            final String encodedPath = FileServletUtil.encodeFileNamePart(request.getParameter("path").replaceAll("\\s", "%20"));
             final URI uri = new URI(encodedPath);
 
             if (!validateAccess(uri,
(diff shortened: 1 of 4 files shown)


tsurdilo:
need this fix for the 7.5.x branch as well and idk which AF branch this translates to

jomarko:
@tsurdilo not sure if I built wrongly the PR but seems to me like still the same behavior, if I create project with name `a b c` and then create there business process, save the process, svg is generated, I try to download svg, however this is not downloaded. Am I doing something wrong please?

ederign:
@tsurdilo can you please take a look at the @jomarko comment?

tsurdilo:
@jomarko @ederign sorry I completely forgot about this one :( 

@jomarko I am fixing issues with designer atm so that won't happen. I think if test in this pr pass ok we are good to go

tsurdilo:
@jomarko look at https://issues.jboss.org/browse/JBPM-6694 again - we talked about this one there.

jomarko:
@tsurdilo I am looking, thanks for notice.

tsurdilo:
@jomarko i'm not sure how to fix this and need some help from @ederign i think here. 
With your example the image does get uploaded fine. The issue then becomes that later on in

JGitFileSystemProvier on lines 1932 to 1934:

if (pathResult.getPathType().equals(NOT_FOUND)) {
            throw new NoSuchFileException(path.toString());
        }

the exception gets thrown for a path which contains %20 (instead of spaces). 
I am not sure how to fix this I think the fix would be to just allow uri encoded strings as path (so decode them before this getPathType() check but I am not sure. So this seems unrelated to this fix specifically but is side-effect of it. WDYT?

tsurdilo:
@ederign another question is why we don't do the same with project names that contain spaces as we do for package names that contain spaces (where they seem to be stripped out from the path). I think that would be the real solution to these types of issues.

jomarko:
@tsurdilo you are probably right that this is new issue, that was hidden because of the AF-934. I filed the new issue [AF-971](https://issues.jboss.org/browse/AF-971)

tsurdilo:
@jomarko thanks! ok i will merge this as it blocks other prs and we can look at the raised issue as part of https://issues.jboss.org/browse/AF-971

PR #121, opened 2017-12-19
[RHDM-217]: Error "Invalid index name [system_ou/plugins]" when using elastic search

adrielparedes:
All "/" inside index names are replaced with "_". 

@jhrcek

### DIFF
--- a/uberfire-extensions/uberfire-metadata/uberfire-metadata-backends/uberfire-metadata-backend-elasticsearch/src/main/java/org/uberfire/ext/metadata/backend/elastic/index/ElasticSearchIndexProvider.java
+++ b/uberfire-extensions/uberfire-metadata/uberfire-metadata-backends/uberfire-metadata-backend-elasticsearch/src/main/java/org/uberfire/ext/metadata/backend/elastic/index/ElasticSearchIndexProvider.java
@@ -143,8 +143,8 @@ public class ElasticSearchIndexProvider implements IndexProvider {
                                           ElasticMetaProperty::getValue,
                                           (mp1, mp2) -> mp2));
 
-        return this.getClient().prepareIndex(clusterId,
-                                             type.toLowerCase()).setSource(document);
+        return this.getClient().prepareIndex(sanitizeIndex(clusterId),
+                                             sanitizeIndex(type)).setSource(document);
     }
 
     @Override
@@ -277,7 +277,7 @@ public class ElasticSearchIndexProvider implements IndexProvider {
                 searchSourceBuilder.size(ELASTICSEARCH_MAX_SIZE);
             }
             return Optional.of(this.getClient()
-                                       .prepareSearch(indicesToLowerCase(indexes).toArray(new String[indexes.size()]))
+                                       .prepareSearch(sanitizeIndexes(indexes).toArray(new String[indexes.size()]))
                                        .setSource(searchSourceBuilder).get());
         } catch (ElasticsearchException e) {
             logger.debug(MessageFormat.format("Can't perform search: {0}",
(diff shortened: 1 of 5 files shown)


uberfire-extensions/uberfire-metadata/uberfire-metadata-backends/uberfire-metadata-backend-elasticsearch/src/main/java/org/uberfire/ext/metadata/backend/elastic/index/ElasticSearchIndexProvider.java
```diff
@@ -313,8 +313,16 @@ private String escape(String s) {
         return sb.toString();
     }
 
-    private List<String> indicesToLowerCase(List<String> indices) {
-        return indices.stream().map(String::toLowerCase).collect(Collectors.toList());
+    protected List<String> prepareIndices(List<String> indices) {
+        return indices.stream()
+                .map(index -> index.toLowerCase().replaceAll("/",
```
jhrcek: You might want to call `.map(this::prepareIndex)` here just to concentrate all index sanitization logic in one place it needs to be extended in the future.

uberfire-extensions/uberfire-metadata/uberfire-metadata-commons-io/src/test/java/org/uberfire/ext/metadata/io/elasticsearch/suite/ElasticSearchTestSuite.java
```diff
@@ -72,8 +72,6 @@
                  "9300")
             .env("xpack.security.enabled",
                  "false")
-            .mountFrom(new File("src/test/resources/elasticsearch.yml").getAbsolutePath())
```
jhrcek: Would it make sense to remove `src/test/resources/elasticsearch.yml` from resources (provided you're not using it anywhere else)?

uberfire-extensions/uberfire-metadata/uberfire-metadata-backends/uberfire-metadata-backend-elasticsearch/src/main/java/org/uberfire/ext/metadata/backend/elastic/index/ElasticSearchIndexProvider.java
```diff
@@ -313,8 +313,16 @@ private String escape(String s) {
...
+                                                             "_"))
+                .collect(Collectors.toList());
+    }
+
+    protected String prepareIndex(String index) {
```
jhrcek: This method name is a bit confusing, because `org.elasticsearch.client.Client` which you're using in this class already has a method like this which does something radically different.

Perhaps you could rename it to `sanitizeString`, `sanitizeIndexName`, `replaceInvalidChars` or something like that.

uberfire-extensions/uberfire-metadata/uberfire-metadata-backends/uberfire-metadata-backend-elasticsearch/src/main/java/org/uberfire/ext/metadata/backend/elastic/index/ElasticSearchIndexProvider.java
```diff
@@ -344,14 +352,14 @@ public void putMapping(String index,
...
+            this.getClient().admin().indices().prepareCreate(prepareIndex(index)).get();
         } catch (ResourceAlreadyExistsException ex) {
             logger.debug("Resource Already exists: " + ex.getMessage());
         }
         Map<String, Object> properties = this.createMappingMap(metaObject.getProperties());
         this.getClient().admin().indices()
-                .preparePutMapping(index.toLowerCase())
-                .setType(type.toLowerCase())
+                .preparePutMapping(prepareIndex(index))
+                .setType(prepareIndex(type))
```
jhrcek: Just being curious: how did you find out that parameters for `setType` also need to be sanitized? Is it really needed?

adrielparedes: @jhrcek I guess that types are not going to have this problem, but in the other hand, I can't ensure this is not going to happen, that is why I decide to sanitize them too

PR #122, opened 2018-01-04
DROOLS-2181: Icons (Edit, Collapse, Delete etc) do not respond to mouse clicks

manstis:
See https://issues.jboss.org/browse/DROOLS-2181

The problem was that two instances of the underlying View widget were being created; one for the actual editor (as seen by the User) and another hidden instance used to drive Editor permissions. The two instances both responded to various CDI events triggered from User interactions in the editor causing the unexpected behaviour.

An alternative approach was to simply make the Editor ```@ApplicationScoped``` but I thought to fix the cause of the issue (as other editors are subject to permissions by default - in BxMS - and any editor at the User's volition).


manstis:
@dgutierr FYI.. Please approve.

@ederign FYI, as AppFormer lead. Please approve.

PR #126, opened 2018-01-05
Add JaCoCo to AppFormer.

kkufova:
To be able to use JaCoCo properly, it is first needed to increase the version of PowerMock. Otherwise, the `JAASAuthenticationServiceTest` test in the `org.uberfire.backend.server.security` package fails.

The version is increased in this PR: https://github.com/jboss-integration/jboss-integration-platform-bom/pull/414.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,10 +19,9 @@
 
   <modelVersion>4.0.0</modelVersion>
   <parent>
-    <groupId>org.jboss.integration-platform</groupId>
-    <artifactId>jboss-integration-platform-bom</artifactId>
-    <!-- Keep in sync with the parent version of uberfire-bom/pom.xml -->
-    <version>8.0.0.CR3</version>
+    <groupId>org.kie</groupId>
+    <artifactId>kie-parent</artifactId>
+    <version>7.6.0-SNAPSHOT</version>
   </parent>
 
   <groupId>org.uberfire</groupId>


kkufova:
@manstis, could you please have a look?

pom.xml
```diff
@@ -142,6 +142,11 @@
           two versions are being updated on the IP BOM.-->
     <version.com.fasterxml.jackson>2.6.2</version.com.fasterxml.jackson>
     <version.commons-cli>1.3.1</version.commons-cli>
+
+    <!-- JaCoCo plugin configurations: -->
+    <jacoco.haltOnFailure>false</jacoco.haltOnFailure>
+    <jacoco.line.coveredratio.minimum>0.9</jacoco.line.coveredratio.minimum>
+    <version.jacoco.plugin>0.7.9</version.jacoco.plugin>
```
manstis: You need to submit a PR to ```droolsjbpm-build-bootstrap``` to change the version of JaCoCo used. You also do not need to define "halt on failure" or "line covered ration min" here (as they're no different to those in ```droolsjbpm-build-bootstrap```.

See https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/pom.xml#L190

kkufova:
@manstis, I checked the AppFormer's effective POM and it is not dependent on `kie-parent` (`droolsjbpm-build-bootstrap`). For this reason, I think I need to define the version, halt on failure, and line covered ratio here as well; I wanted to make it as similar as possible to the JaCoCo configuration available in `droolsjbpm-build-bootstrap`, so I decided to leave these parameters here as well.

Regarding the "offline instrumentation", to be honest, I don't understand why, but the JaCoCo does not work in AppFormer with the "on the fly instrumentation" (I tried both configurations). I used the same instrumentation as in the `droolsjbpm-build-bootstrap`.

I am aware of the PR created by @psiroky, but since he is no longer working here, I don't know who to ask about the safety of increasing the PowerMock version. Do you have any idea who could help me with this issue?

Please let me know what you think.

manstis:
@kkufova Thanks for the explanations.

There's a few things I'd recommend:-

1) AppFormer can now inherit from ```droolsjbpm-build-bootstrap``` as AppFormer is now built as part of the larger KIE "family" (historically AppFormer - previously named UberFire - had its own release schedule and was not built as part of the KIE "family").

2) I'd then suggest changing the PowerMock version in ```droolsjbpm-build-bootstrap``` to that you require for the JaCoCo plugin. Perform a "full downstream build" to flush out any issues of using the newer version.

3) I'd finally recommend, assuming 1 and 2 have gone OK, merging this (modified, following 1 and 2) JaCoCo PR!

@ederign Do you concur with this plan?

kkufova:

kkufova:
@manstis, thank you for your help! I hope I understood you correctly. I changed the parent of AppFormer to `kie-parent`, so it can use JaCoCo defined in `droolsjbpm-build-bootstrap`. The PowerMock version is defined [here](https://github.com/jboss-integration/jboss-integration-platform-bom/blob/master/pom.xml#L329), so I believe the other [PR](https://github.com/jboss-integration/jboss-integration-platform-bom/pull/414) is correctly placed. Let's wait for the full downstream build.

porcelli:
we need @mbarkley to look into this

kkufova:
The JaCoCo thing got really out of hand here. :-)

@mbarkley, thank you for looking into this!

As I have zero experience with this and @mbarkley already has the issue in progress, I suggest closing these PRs (this one, https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/673, and https://github.com/jboss-integration/jboss-integration-platform-bom/pull/414). If @mbarkley decides to stop working on this in the future, I can take over the issue and try to finish it. @manstis, do you agree?

manstis:
@kkufova Fine with me - it doesn't affect anything I am working on :-)

PR #129, opened 2018-01-08
DROOLS-1946 Change implementation of moving rows

jomarko:
@manstis I changed `org.uberfire.ext.wires.core.grids.client.model.impl.BaseGridData#moveRows` method a little bit and added couple of tests. However one test[1] is still failing. If I keep old implementation of `org.uberfire.ext.wires.core.grids.client.model.impl.BaseGridData#moveRows`  multiple added test will fail, could you please help?

[1]  
`org.uberfire.ext.wires.core.grids.client.model.impl.GridRowsMergedTest#testMergedBlock_MoveRowsUp_Indexes3and4to1`

Ensemble with:
- https://github.com/kiegroup/kie-wb-common/pull/1339

### DIFF
--- a/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/model/impl/BaseGridData.java
+++ b/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/model/impl/BaseGridData.java
@@ -236,14 +236,7 @@ public class BaseGridData implements GridData {
         }
 
         this.rows.removeAll(rows);
-
-        if (index < oldBlockStart) {
-            this.rows.addAll(index,
-                             rows);
-        } else if (index > oldBlockStart) {
-            this.rows.addAll(index - rows.size() + 1,
-                             rows);
-        }
+        this.rows.addAll(index, rows);
 
         final Range oldBlockExtent = new Range(oldBlockStart,
                                                oldBlockEnd);
--- a/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/model/impl/BaseGridDataIndexManager.java
+++ b/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/model/impl/BaseGridDataIndexManager.java
@@ -614,6 +614,9 @@ public class BaseGridDataIndexManager {
                 updateMergeMetaData(oldBlockEnd + 1,
                                     _columnIndex);
             }
+            if (oldBlockEnd == rows.size() - 1) {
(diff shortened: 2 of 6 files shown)


manstis:
Your test has indeed found a bug. Congratulations!

The last row is not being re-indexed correctly when the block being moved includes the last row.

Please add this to your PR at L#617 [here](https://github.com/kiegroup/appformer/blob/master/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/model/impl/BaseGridDataIndexManager.java#L617).
```
if (oldBlockEnd == rows.size() - 1) {
    updateMergeMetaData(oldBlockEnd, _columnIndex);
}
```
Your test is also incorrect. Row 4 does not contain any merged cells and therefore https://github.com/kiegroup/appformer/pull/129/files#diff-b2813814ceea779b0c03301b2e999a0aR1974 should read ```true, true, true, true, false```.

manstis:
... and you're sure you've checked this with the Guided Decision Table Editor with **collapsed** rows (not just merged, but merged and collapsed?) When a cell spans multiple rows you can collapse the cell into a single row and then drag the single row to a new position.

jomarko:
@manstis looks great, sure, I will increase the coverage also.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/model/impl/BaseGridData.java
```diff
@@ -236,14 +236,7 @@ public void moveRowsTo(final int index,
...
         this.rows.removeAll(rows);
-
-        if (index < oldBlockStart) {
-            this.rows.addAll(index,
-                             rows);
-        } else if (index > oldBlockStart) {
-            this.rows.addAll(index - rows.size() + 1,
-                             rows);
-        }
+        this.rows.addAll(index, rows);
```
jomarko: Important change here

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/model/impl/BaseGridDataIndexManager.java
```diff
@@ -614,6 +614,9 @@ public void onMoveRows(final List<GridRow> rowsMoved,
                 updateMergeMetaData(oldBlockEnd + 1,
                                     _columnIndex);
             }
+            if (oldBlockEnd == rows.size() - 1) {
+                updateMergeMetaData(oldBlockEnd, _columnIndex);
+            }
```
jomarko: Important change here.

manstis:
@jomarko Those tests should be enough; thanks for looking (I also ran ```drools-wb-webapp``` yesterday with your changes in this PR and the Guided Decision Table editor, with collapsed rows, worked fine)... so happy to merge.

jomarko:
@manstis thanks for double check by running locally, PR should be ready for merge now.

PR #130, opened 2018-01-09
AF-927: [Layout Editor] Expose all the Displayer types as individual components

dgutierr:
@kkufova Ready for review => https://issues.jboss.org/secure/attachment/12429675/12429675_cms-displayerdragcomponents.mp4

### DIFF
--- a/dashbuilder/dashbuilder-client/dashbuilder-cms-client/src/main/java/org/dashbuilder/client/cms/layout/editor/PerspectiveDragComponent.java
+++ b/dashbuilder/dashbuilder-client/dashbuilder-cms-client/src/main/java/org/dashbuilder/client/cms/layout/editor/PerspectiveDragComponent.java
@@ -52,6 +52,11 @@ public class PerspectiveDragComponent implements PerspectiveEditorCoreComponent,
     }
 
     @Override
+    public String getDragComponentIconClass() {
+        return "fa fa-file-o";
+    }
+
+    @Override
     public IsWidget getPreviewWidget(RenderingContext ctx) {
         return getShowWidget(ctx);
     }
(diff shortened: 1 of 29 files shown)


kkufova:
@dgutierr, there is one test failure in `PerspectiveEditorPresenterTest.testInitDragComponentGroups` (lines 132 and 133). Could you please take a look? It should be enough to switch the values.

dgutierr:
@kkufova I've fixed so. Thanx!

kkufova:
@dgutierr, everything works nice, but there is one thing I am not quite sure about -- and that is that adding anything from the **Charts** section actually opens the very same modal, but only with a different tab opened. The same thing happens with **Filters**, but with a different subtab is opened. This feels a little bit redundant, especially because the list of charts is quite long. Two solutions come to my mind:
1. A modal that does not allow the user to switch between different charts could open for each of the items;
2. these two sections, **Charts** and **Filters**, could be merged into only one category (for example **Displayers**) with two items; Charts and Filters. Then, maybe these two items could open a modal window that does not allow switching between them.

WDYT?

dgutierr:
@kkufova I see your point. not sure about it though. I do think it's better to let the user change the displayer type once the modal is opened. This is really important when editing an already dropped displayer, which you can edit and change its type without having to drop it and create a new one. 

Also, the palette elements are only a shortcut to the types in the modal. See at it as an usability enhancement.   

As for the palette categories I'm not sure what's the best categorization. Charts and Filters sound good to me, but I'd like to know what others think, specially uxd.

kkufova:
@dgutierr, I agree, it's true that it is a nice feature to be able to change the type of some chart after it is already dropped. Let's discuss the categorization with the UX team then.

kkufova:
@dgutierr, could we implement the changes suggested in [AF-927](https://issues.jboss.org/browse/AF-927) (icons, categorization) before merging this?

dgutierr:
@kkufova I'm on it... :-)

dgutierr:
@kkufova  Done = https://issues.jboss.org/secure/attachment/12430043/12430043_cms-displayerdragcomponents2.mp4

Ready for the last review I think.

kkufova:
@dgutierr, works well, thanks!

dgutierr:
@ederign Can you review it as well please? Thanx!

dgutierr:
@ederign Can you merge please?

PR #132, opened 2018-01-09
AF-497:Latest Version selector set to "Version -1"

jsoltes:
This is a reproducer and a fix for AF-497: Latest Version selector set to "Version -1" after rename and save. The problem was that the VersionRecordManager was not being updated after the rename of the asset, so when the versions were reloaded on the save event, the title was updated to -1. The fix is to add VersionRecordManager.init into the BaseEditor class.


uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/test/java/org/uberfire/ext/editor/commons/client/BaseEditorRenameTest.java
```diff
@@ -0,0 +1,94 @@
+package org.uberfire.ext.editor.commons.client;
```
manstis: License please.

jsoltes: you were faster than me :)

uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/test/java/org/uberfire/ext/editor/commons/client/BaseEditorRenameTest.java
```diff
@@ -0,0 +1,110 @@
...
+    @Test
+    public void testVersionRecordManagerIsInitializedOnRename() {
+        //clear the interaction on versionRecordManager from the baseEditor init method
+        Mockito.reset(versionRecordManager);
+
+        when(versionRecordManager.getCurrentPath()).thenReturn(observablePath);
+        baseEditor.onRename();
+
+        verify(versionRecordManager).init(any(),
+                                          any(),
```
manstis: Wouldn't it be nice to check the path was ```observablePath``` rather than _any_ path?

ederign: @jsoltes could you please fix this?

jsoltes:
@manstis @ederign updated

PR #134, opened 2018-01-11
[7.5.x] [Content Management] HTML panel causes errors when saving & when adding page to navigation tree

karreiro:
See: https://issues.jboss.org/browse/RHDM-219

---

demo

---

Regarding the [ignored exception](https://github.com/Voog/wysihtml/blob/26a85fc0e51bde077c23bcdcbdc0549f34341e0c/lib/rangy/rangy-core.js#L1318), it was introduced by the `wysihtml` [version update](https://github.com/kiegroup/appformer/commit/24423b31654e74739f2e2af0c0c3f0f26eecce34#diff-620e02db07d901ae62cd35ca89329df4). Since everything still works as expected and the lib is in a beta version, I think that we can wait for a more stable API and ignore the exception for now.

The fix refers to loading and destroying the `wysihtml` editor everytime we show and hide the modal, respectively.

---

Part of an ensemble:
- https://github.com/kiegroup/appformer/pull/134
- https://github.com/kiegroup/kie-wb-distributions/pull/668

### DIFF
--- a/uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/htmleditor/HtmlEditorView.java
+++ b/uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/htmleditor/HtmlEditorView.java
@@ -23,7 +23,6 @@ import javax.inject.Inject;
 
 import com.google.gwt.core.client.JavaScriptObject;
 import com.google.gwt.core.client.Scheduler;
-import com.google.gwt.user.client.Event;
 import com.google.gwt.user.client.Window;
 import com.google.gwt.user.client.ui.Widget;
 import org.jboss.errai.common.client.dom.Anchor;
@@ -36,8 +35,6 @@ import org.jboss.errai.common.client.ui.ElementWrapperWidget;
 import org.jboss.errai.ui.client.local.api.IsElement;
 import org.jboss.errai.ui.client.local.spi.TranslationService;
 import org.jboss.errai.ui.shared.api.annotations.DataField;
-import org.jboss.errai.ui.shared.api.annotations.EventHandler;
-import org.jboss.errai.ui.shared.api.annotations.ForEvent;
 import org.jboss.errai.ui.shared.api.annotations.Templated;
 import org.uberfire.client.workbench.docks.UberfireDocksInteractionEvent;
 import org.uberfire.ext.editor.commons.client.EditorTitle;
@@ -182,6 +179,7 @@ public class HtmlEditorView implements HtmlEditorPresenter.View,
     private EditorTitle title;
     private boolean loaded = false;
     private JavaScriptObject jsEditor;
+    private String identifier;
 
     @Inject
     public HtmlEditorView(final TranslationService translationService,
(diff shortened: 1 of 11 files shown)


jhrcek:

jhrcek:
Error in Full downstream build

```
08:01:40 FATAL: Remote call on ci-rhos-rhel7-xlarge-1018 failed. The channel is closing down or has closed down
```

Trying again. 

karreiro:
Thank you for the review, @jhrcek. All changes were applied.

uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/htmleditor/HtmlEditorView.java
```diff
@@ -231,19 +229,30 @@ private void handleColorSelection(MouseEvent event, Button button, Input input)
...
 
+            identifier = String.valueOf(System.currentTimeMillis());
             loaded = true;
+
+            configureScreenComponents(getEditorId(), getToolbarId());
+            loadEditor(getEditorId(), getToolbarId());
         }
     }
 
+    private String getToolbarId() {
```
jhrcek: @karreiro thanks for the change :) Could you please also remove these helper methods? Now that destroyEditor no longer uses them, they're only used in `load()` method. I think it would be best to switch back to previous implementation which was using final string variables `identifier`, `editorId` and `toolbarId` inside `load()` method, minimizing their scope and changes introduced as well..

karreiro: @jhrcek Thanks for the suggestion. But I prefer to keep it as is in this case. Separating the "id construction logic" from the "load logic" is a good approach. Besides, (considering the open/closed principle) it’s easier to reuse the identifier in future cases.

jhrcek: Ok let's leave it as is. FYI still waiting for full downstream build to verify locally..

karreiro:
@jhrcek I think that the full downstream passes. But, I did a new commit removing the `destroyEditor` parameters.

uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/htmleditor/HtmlEditorView.java
karreiro: @jhrcek I think that the full downstream passes. But, I did a new commit removing the destroyEditor parameters.

jhrcek:
@karreiro So after local testing I can confirm, that the first issue from that jira was fixed (no more modals with WRONG_DOCUMENT_ERR). But the 2nd error still appears. See steps to reproduce in this video
[ErrorAfterAddingPageToNewGroup.zip](https://github.com/kiegroup/appformer/files/1626127/ErrorAfterAddingPageToNewGroup.zip)

karreiro:
Thank you, @jhrcek. I just applied the fix.

jhrcek:

karreiro:
The broken test is not related to this PR:
```org.kie.workbench.common.services.refactoring.backend.server.drl.IndexDrlLHSTypeExpressionFieldCombinedTest.testIndexDrlLHSTypeExpressionField`

karreiro:

jhrcek:
@karreiro thanks. I'm also trying to build this locally, will hopefully provide feedback within few hours.

karreiro:

PR #147, opened 2018-01-24
[RHDM-306, RHDM-307, RHDM-309]

adrielparedes:
With @paulovmr fixed the issues below:

https://issues.jboss.org/browse/RHDM-306
https://issues.jboss.org/browse/RHDM-307
https://issues.jboss.org/browse/RHDM-309

There was a problem with colons in urls (it seems it's an ES problem with query strings). So now every project have their own assets.

It also fixes a problem with an error that appeared when creating some assets. It was a sort that contains a null field. In lucene that enables the "sort by doc number", but in ES that is by default so we omitted those sort field that are null.

@ederign could you review?


ederign:

ederign:
Test fails are not related.

PR #153, opened 2018-02-01
upgrade to next development version

kiereleaseuser:


manstis:
I don't follow AF's versioning however wonder what happened to 2.3?

mbiarnes:
@manstis the 7.6.0.Final will have the uberfire and dashbuilder 2.3.0.Final

PR #154, opened 2018-02-02
DROOLS-2218: Log NPE during weld cleanup

jomarko:
Running tests which use the CDITestSetup on jenkins sometimes cause NPE in the weld cleanup. Such class and NPE:
- org.drools.workbench.screens.dtablexls.backend.server.DecisionTableXLSServiceImplCDITest
- https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/drools-wb-pullrequests-7.5.x/46/testReport/org.drools.workbench.screens.dtablexls.backend.server/DecisionTableXLSServiceImplCDITest/testValidateColumnsNotInStandardOrder/

@manstis please have a look,
Ensemble with:
kiegroup/drools-wb#794

### DIFF
--- a/uberfire-test-utils/src/main/java/org/guvnor/test/CDITestSetup.java
+++ b/uberfire-test-utils/src/main/java/org/guvnor/test/CDITestSetup.java
@@ -22,6 +22,7 @@ import javax.enterprise.inject.spi.BeanManager;
 
 import org.jboss.weld.environment.se.Weld;
 import org.jboss.weld.environment.se.WeldContainer;
+import org.slf4j.LoggerFactory;
 import org.uberfire.backend.server.util.Paths;
 import org.uberfire.java.nio.fs.file.SimpleFileSystemProvider;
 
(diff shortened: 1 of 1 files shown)


uberfire-test-utils/src/main/java/org/guvnor/test/CDITestSetup.java
```diff
@@ -47,7 +50,12 @@ public void setUp() throws Exception {
 
     public void cleanup() {
         if (weld != null) {
-            weld.shutdown();
+            try {
+                weld.shutdown();
+            } catch (NullPointerException npeException) {
+                Logger.getLogger(CDITestSetup.class.getName())
```
jhrcek: Just out of curiosity: is there a reason to use java util logging (as opposed to slf4j used in most other places)?

jomarko: I would say no, will update. Thanks.

jomarko:
@manstis @jhrcek let me know, if anything else I could improve.

manstis:
@jomarko I'm tempted to suggest catching ```Exception``` in case there's similar weird scenarios in WELD cleanup... but perhaps it doesn't hurt only catching ```NullPointerException``` for now and see what the future brings!

jomarko:
@manstis well, I was thinking about catching most general `Exception` but I think it is better to catch this specific `NPE`, because we know this occurs. We don't know other exceptions that could occur, so I believe it is better to get fail when something new occur, then investigate and probably catch also this new kind of exception.

PR #161, opened 2018-02-05
Fix UI issues when removing space with multiple projects.

mbarkley:
Fixes an issue introduced by AF-648 (Project Oriented).

### DIFF
--- a/uberfire-project/uberfire-project-api/src/main/java/org/guvnor/common/services/project/context/WorkspaceProjectContextChangeEvent.java
+++ b/uberfire-project/uberfire-project-api/src/main/java/org/guvnor/common/services/project/context/WorkspaceProjectContextChangeEvent.java
@@ -15,6 +15,8 @@
  */
 package org.guvnor.common.services.project.context;
 
+import java.util.Objects;
+
 import org.guvnor.common.services.project.model.Module;
 import org.guvnor.common.services.project.model.Package;
 import org.guvnor.common.services.project.model.WorkspaceProject;
@@ -67,7 +69,7 @@ public class WorkspaceProjectContextChangeEvent {
     public WorkspaceProjectContextChangeEvent(final WorkspaceProject workspaceProject,
                                               final Module module,
                                               final Package pkg) {
-        this.ou = workspaceProject.getOrganizationalUnit();
+        this.ou = workspaceProject != null ? workspaceProject.getOrganizationalUnit() : null;
         this.workspaceProject = workspaceProject;
         this.module = module;
         this.pkg = pkg;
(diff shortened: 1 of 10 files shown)


mbarkley:
Related PR: https://github.com/kiegroup/kie-wb-common/pull/1405

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/repositories/RepositoryServiceImpl.java
```diff
@@ -322,8 +323,9 @@ public void removeRepositories(Space space, Set<String> aliases) {
                                                                     .format("The given space [%s] does not correspond to any known organizational unit.",
                                                                             space.getName())));
 
+            List<Repository> removedRepos = new ArrayList<>();
```
jhrcek: Could you please explain what's the purpose of this `removedRepos` list? You seem to be accumulating repos in it while they're being deleted, but then you don't do anything with it..
Shouldn't you fire some "repo deleted" notifications like you're doing in the method above?

This would probably deserve a unit test

paulovmr: Same comment here.

uberfire-project/uberfire-project-api/src/main/java/org/guvnor/common/services/project/context/WorkspaceProjectContextChangeEvent.java
```diff
@@ -88,4 +88,52 @@ public Module getModule() {
...
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj)
+            return true;
+        if (obj == null)
+            return false;
+        if (getClass() != obj.getClass())
+            return false;
+        WorkspaceProjectContextChangeEvent other = (WorkspaceProjectContextChangeEvent) obj;
+        if (module == null) {
```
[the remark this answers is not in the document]
mbarkley: Yeah, what's there now is pretty ugly (was generated by Eclipse). I can change it.

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/repositories/RepositoryServiceImpl.java
mbarkley: Oops. I had meant to remove this before sending the PR. I initially thought about firing a new event type containing the list of removed repositories so that batch deletions could be observed. Ultimately I decided it wasn't needed (yet, at least), since multiple repositories are only removed at once when removing a space (and removing a space fires an event already).

I could make a new event signalling a batch deletion of repositories, or I could just removed this. WDYT? Personally, I'd prefer just removing it and not adding a new event until it's clear that we need it.

mbarkley:
@jhrcek @paulovmr I've removed the unused code and added documentation to the interface the `removeRepositories` does not fire CDI events. I've also added a test for `removeOrganizationalUnit` on the `OrganizatoinalUnitService` that ensures that an OU's repositories are removed and that the fired event contains all the deleted repositories so that there is a way to observe there removal if required.

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/repositories/RepositoryServiceImpl.java
jhrcek: @mbarkley I agree with removal of the notification for deletion of multiple repos. I suppose that deletion of space should be done by (power) users rather rarely and they should know what they're doing.

PR #164, opened 2018-02-06
Fix for running showcase GWT dev mode using IntelliJ

cristianonicolai:


cristianonicolai:
@ederign not sure if it is only me having issues to run the showcase using IntelliJ 2016.2 but I had to do these changes in order to get it running.

PR #165, opened 2018-02-06
AF-986: clone via rest has to set mirror to false

porcelli:


porcelli:
@jhrcek the reported jira is not exactly related to this... check RHDM jira for details

but this fix seems to be not complete, as I don't know why it's not storing that repo is not a mirror and keep syncing with `mirror`

PR #166, opened 2018-02-06
DROOLS-2303: [DMN Designer] Add column to whole row outline

manstis:
See https://issues.jboss.org/browse/DROOLS-2303

@jomarko This fix also applies to Guided Decision Tables where it too was possible to select a "row" and then insert a new column that was then not included in the selected row.


manstis:
Jenkins please retest this.

PR #169, opened 2018-02-07
DROOLS-2303: post review actions

jomarko:
@manstis please have a look

Sorry for the amount of changed lines, but in final I removed ~500 "test setup lines" from `GridCellSelectionsTest`


jomarko:
@manstis jenkins says green.

PR #175, opened 2018-02-08
JBPM-6659 - New layout presenter for handling master/detail navigatio…

cristianonicolai:
…n across screens

### DIFF
--- a/uberfire-showcase/uberfire-webapp/src/main/java/org/uberfire/client/perspectives/MultiScreenPerspective.java
+++ b/uberfire-showcase/uberfire-webapp/src/main/java/org/uberfire/client/perspectives/MultiScreenPerspective.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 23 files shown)


cristianonicolai:
@nmirasch please take a look. You can see the new layout using the showcase webapp -> Perspectives -> Multi Screen Perspective menu.

bsremac:
@cristianonicolai These buttons _Open TodoListScreen_ and _Open side panel TodoListScreen_, what screens they should correspond to in the context of existing jbpm-wb perspectives?
I assume that _Multi Screen_ drop-down would correspond to the new _Manage_ selector, right?

cristianonicolai:
@bsremac _Open TodoListScreen_ represents a new detail page, it should open taking the entire screen.
_Open side panel TodoListScreen_ represents the current detail pages, in a side panel but using this new layout.
Correct for the _Manage_ selector.

cristianonicolai:
@ederign the side panel was just an interim solution. We're moving to master detail navigation in the manage section which basically stacks screens on top of each other as shown in the _Open TodoListScreen_ example. In terms of design, these are not targetting an editor like approach. Although we support the different layout for the buttons, we will only have one or two actions.
See design for reference: https://rawgit.com/lclay2/bxms-uxd-sandbox/master/design/Manage-section/process_details_existing.html

ederign:
@cristianonicolai thanks for the clarification!

PR #176, opened 2018-02-08
Update Wires WildFly dependency

cristianonicolai:


cristianonicolai:
@ederign this one seems to work with both dependencies, so just updating here to keep consistency with uberfire showcase setup.

PR #181, opened 2018-02-09
Close busy popup when view closed to prevent issues from ERRAI-1095

mbarkley:
[ERRAI-1095](https://issues.jboss.org/projects/ERRAI/issues/ERRAI-1095) prevents callbacks from invoking on Errai callers after the enclosing bean is destroyed. A possible side-effect is that some beans can be closed before they hide popups, causing the popups to remain indefinitely. This prevents that issue for loading popups created with the `BusyIndicatorView`.


mbarkley:
@manstis @ederign Let me know if there are other beans that you think could also use this enhancement.

PR #182, opened 2018-02-09
AF-1007: Workbench *.cache.js files still not properly cached, even w…

mbarkley:
…hen cache header filter configured


mbarkley:
At least in WildFly 10, setting the pragma header to "cache" stops undertow from overriding it.

mbarkley:
@ederign Mind reviewing this?

PR #185, opened 2018-02-13
RHDM-382: Close busy popup when view closed to prevent issues from ER…

mbarkley:
…RAI-1095

### DIFF
--- a/uberfire-extensions/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/common/BusyIndicatorView.java
+++ b/uberfire-extensions/uberfire-widgets/uberfire-widgets-commons/src/main/java/org/uberfire/ext/widgets/common/client/common/BusyIndicatorView.java
@@ -15,6 +15,7 @@
  */
 package org.uberfire.ext.widgets.common.client.common;
 
+import javax.annotation.PreDestroy;
 import javax.enterprise.context.Dependent;
 
 @Dependent
@@ -25,7 +26,13 @@ public class BusyIndicatorView implements HasBusyIndicator {
         BusyPopup.showMessage(message);
     }
 
+    /**
+     * This method is invoked by the container when this bean is destroyed,
+     * so that loading popups are guaranteed to close when an enclosing
+     * bean is destroyed.
+     */
     @Override
+    @PreDestroy
     public void hideBusyIndicator() {
         BusyPopup.close();
     }


mbarkley:
@jhrcek I've tested this fix myself, but would you mind verifying it?

jhrcek:

jhrcek:
@mbarkley So I compiled appformer (PR branch) + kie-wb-distributions (7.5.x) locally with sourcemaps enabled. I double checked the fix from this PR ends up in the final war. After playing with the debugger I confirm the problem seems to come from [ProjectScreen](https://github.com/kiegroup/kie-wb-common/blob/7.5.x/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/ProjectScreen.java#L204) which has rather complex asset loading logic which is executed [periodically every 2 seconds](https://github.com/kiegroup/kie-wb-common/blob/7.5.x/kie-wb-common-screens/kie-wb-common-library/kie-wb-common-library-client/src/main/java/org/kie/workbench/common/screens/library/client/screens/ProjectScreen.java#L298-L305) using scheduler. So it seems that closing the popup in PreDestroy won't help, because ProjectScreen is still alive when the popup appears. Instead we should probably call  `busyIndicatorView.hideBusyIndicator();` at appropriate place in ProjectScreen, when the periodic asset loading process is interrupted by the project being deleted. But I've no idea what would be the appropriate place to call that. Perhaps @paulovmr could help?

manstis:
@mbarkley Completely unrelated - sort of - IIRC you was asking what you could do (in general) on AF's IRC channel yesterday? Well the code @jhrcek points to is, IMO, really nasty. The Project Screen polls the backend periodically fetching the asset list. If the content "changes" the code assumes indexing is incomplete and keeps polling. No change and it assumes indexing is complete... Personally, I'd prefer for the indexing to push "status" events (started, update, complete) to the client and remove the polling.. WDYT?

jhrcek:
Sorry, I forgot to emphasize that this is a regression since last build (CR2). Before CR3 I never saw this failure (and I double-checked manually with CR2 to confirm that), whereas in CR3 it's reproducible 100%. I'm not sure what changed, though.

Although I agree the current implementation is complex, I'd consider @manstis suggestion to be high risk refactoring for which there's no room within prod. schedule. Also the nice thing about the current implementation is that assets are loaded incrementally into the view. Would that still be the case with the "push" events?

manstis:
@jhrcek I agree.. I did not mean as part of this JIRA/PR but something to consider.

manstis:
@jhrcek If we pushed events during indexing; i.e. as each file is indexed (or via some batching process) then it'd work exactly the same as it does now.

paulovmr:
@manstis @jhrcek I agree, push events would improve a lot this screen performance and code. Even using events it would be possible to load the screen incrementally. Also agree it is something we should do on master, not on this branch unfortunately.

CC'ing @Rikkola as he played more with indexing and pagination on library. Maybe this is something we could improve during this stabilization period, after the project oriented big merge.

Now for this PR scope, @mbarkley if you need help on this search/debugging ping me on IRC. But I guess you know the same as me, maybe more, about that class. Also, no matter what solution you find, I think this `@PreDestroy` should remain to prevent other issues.

mbarkley:
@manstis I also think that would be a great change on master.

@jhrcek I was able to reproduce it again with the war from the downstream build. I think I may have been waiting too long before deleting before (hence the asset reloading may have finished already). Looking into this more now.

> whereas in CR3 it's reproducible 100%. I'm not sure what changed, though.

[ERRAI-1095](https://issues.jboss.org/projects/ERRAI/issues/ERRAI-1095) prevents callbacks to Errai RPC services from being invoked after a bean is destroyed. This was needed to fix a UI issue that was occurring throughout the workbench (a popup message about looking up `@Templated` beans that aren't in service). As a side-effect, some callbacks that used to remove popups after the calling screen closed now don't fire.

mbarkley:
@jhrcek After playing around, it seems like this commit introduces a timing issue. If the project deletes before the asset loading callback is invoked, everything is fine. If the callback is invoked but the scheduled reload hasn't happened, then there is a problem.

The fix is to ensure the scheduled reload is cancelled when the bean is destroyed. I should have another commit soon.

mbarkley:
Added a PR to kie-wb-common: https://github.com/kiegroup/kie-wb-common/pull/1425

mbarkley:
Jenkins, please execute full downstream build.

mbarkley:
Jenkins, please execute full downstream build.

Rikkola:
Nothing in the recent changes that would help this. Like @manstis mentioned the events for the index status would fix this ( and the randomly failing indexing tests we have ). 

Not sure where we could plug the events in. Currently we fire an event that notifies indexing has finished, but what it really does is say "our side of the code has done all it can", indexing still continues to run for X amount of time in the background.

PR #186, opened 2018-02-14
update GitHub URLs

michpetrov:
Some URLs are pointing to the old uberfire repo and leading nowhere, updated to point in here.


ederign:
Thanks @michpetrov

PR #188, opened 2018-02-15
AF-610: [Layout editor] CSS edition

dgutierr:
@kkufova @pefernan You can already start looking at this very initial version => https://issues.jboss.org/secure/attachment/12431287/12431287_cms-property-editor.mp4

Any issues found are welcome :-)

P.S: related PR https://github.com/kiegroup/kie-wb-common/pull/1429


dgutierr:
@ederign Can you please review it as well?

dgutierr:
I've also implemented AF-904, which basically consists of disabling the "Screen" component.

kkufova:

kkufova:
@dgutierr, first of all, great job! Overall, I think it looks good and works really well. :-) There are few things I would like to ask you:

1. Firstly, I noticed that if you make any settings directly in the **Edit HTML Component** modal window (for example change the alignment), the alignment then cannot be changed in the Properties panel (and vice versa). Would it be possible to deal with this in the way that if something is set in one place, it can be later changed in the other?

2. One tiny nitpick regarding texts: I noticed that there is no difference between the weight options Bold and Bolder. Maybe we can remove one of them so the three available options would be e.g. None, Lighter, and Bolder?

3. What is the difference between Margin and Padding for elements other than a page (such as HTML components and Reporting components)? For these components, both of the panels act identically, so I think one of them can be disabled without losing any functionality.

4. Lastly, I noticed that managing pages (creating and deleting them) is quite buggy in this PR. I tried to record it here: https://issues.jboss.org/secure/attachment/12431388/PR188-bug1.webm. Could you please take a look at this? If you need more descriptive steps to reproduce, let me know. I basically just created a page, played with the properties a little bit, and then started to create and delete other pages.

Also, let's consider [AF-904](https://issues.jboss.org/browse/AF-904) done.

dgutierr:
@kkufova 

1. Yes, the component modal settings always takes precedence over the left docked panel properties. In the near future though, we should move the more settings we can from modal config to the docked panel. This is true for all the components types (charts, forms fields, etc...) including the HTML one. Anyway, to be honest, I do not think this is critical short term.

2. I removed the "Bold" option :-)

3. There is a difference. See video https://issues.jboss.org/secure/attachment/12431390/cms-margin-padding.mp4  I'd keep margin/padding as it is right now.

4. It looks like there are issues with the git file life cycle notifications. The callbacks defined  [here](https://github.com/kiegroup/appformer/blob/master/uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/BaseEditor.java#L342) not invoked. Can you double check please whether this is a general workbench issue or something local in the CMS perspective?

kkufova:
@dgutierr, thanks for the info!
1. I will file this so we won't forget about it in the future.
2. Great, thank you!
3. You are completely right, my apologies. I would leave it as it is.
4. Reported: [RHBA-398](https://issues.jboss.org/browse/RHBA-398).

dgutierr:
@kkufova Thanks Klara

I added more fine control over the layout elements. Column's nested components are supported now. So we can have single component columns which we just called as "Column n" or multiple components columns which we enumerate in the dropdown as "Column n>Component m". Please take a 2nd look.

As for the panel's height/width you are right, the settings are applied at component's level not at column level and imho I think it's ok.  Notice, the whole layout is bootstrap based so we need to be careful about what settings the user can override. If width/height might actually cause confusion we could just remove them.

kkufova:
@dgutierr, could you please rebase both repositories (`appformer` and `kie-wb-common`) on `master`? There were some changes that are causing this PR to fail ([this](https://github.com/kiegroup/appformer/pull/195) and [this](https://github.com/kiegroup/kie-wb-common/pull/1447) to be exact). Rebasing fixes the problem. I am just building this PR locally and I'll take one final look after the build finishes.

dgutierr:
@kkufova Done :-)

kkufova:
@dgutierr, I really like the solution for column's nested components, it works well, thanks! There is only one small thing that could be improved; notice this structure:

screenshot from 2018-02-26 12-53-50

-> Although a row always contains "nested" columns, there is still a possibility to manage the whole row (Row 1, Row 2). However, there is no such possibility for columns with nested components: in the above example, I am missing the Row 1 > Column 2 option (but we have Row 1 > Column 1 and Row 1 > Column 3). What do you think?

Regarding the component/column height/width; to be honest, I still don't know what would be the best solution, but let's leave it is as it is for the moment, as the current implementation works nicely and it is possible to set basically everything I can think of anyway. :-)

Also, I can confirm that the build does no longer fail because of the OOM errors!

dgutierr:
@kkufova good question :-)   The reason is that columns are not supported. We do only support rows and components. I'm currently naming there single component columns as "Column a" whereas multiple component columns are named as "Column a>Component n...m".  

We have two options:

1.- Add support for column properties.
2.- Rename single component columns as "Column a > Component n"

I'd go for 2, at least for now.. so in your example above we'd have the following elements listed:

Page
Row 1
Row 1 > Column 1> Component 1
Row 1 > Column 2 > Component 1
Row 1 > Column 2 > Component 2
Row 1 > Column 3 > Component 3
Row 2
Row 2 > Column 1> Component 1

wdyt?

kkufova:
@dgutierr, I see! In this case, I agree with you; for now, we can go for option 2, and I can file a Jira for adding support for single columns and we can address the issue in the future. Do you agree?

dgutierr:
@kkufova Sounds great!

uberfire-extensions/uberfire-layout-editor/uberfire-layout-editor-api/src/main/java/org/uberfire/ext/layout/editor/api/css/CssProperty.java
```diff
@@ -0,0 +1,111 @@
...
+    PADDING_BOTTOM(LENGTH),
+    PADDING_LEFT(LENGTH),
+    PADDING_RIGHT(LENGTH),
+
+    TEXT_ALIGN(LEFT, CENTER, RIGHT),
+    TEXT_DECORATION(NONE, UNDERLINE, OVERLINE, LINE_THROUGH/*, BLINK*/),
+    COLOR(CssValueType.COLOR),
+
+    FONT_SIZE(XX_SMALL, X_SMALL, SMALL, MEDIUM, LARGE, X_LARGE, XX_LARGE),
+    FONT_WEIGHT(NORMAL, /*BOLD,*/ BOLDER, LIGHTER/*, WEIGHT_100, WEIGHT_200, WEIGHT_300, WEIGHT_400, WEIGHT_500, WEIGHT_600, WEIGHT_700, WEIGHT_800, WEIGHT_900*/);
```
ederign: should we remove the commented options?

uberfire-extensions/uberfire-layout-editor/uberfire-layout-editor-client/src/main/java/org/uberfire/ext/layout/editor/client/components/columns/ComponentColumn.java
```diff
@@ -94,10 +109,81 @@ public void init(String parentId,
...
+
+    @Override
+    public boolean isSelected() {
+        return selected;
+    }
+
+    @Override
+    public void setSelected(boolean flag) {
+        if (isSelectable()) {
+            if (flag) {
```
ederign: if (selected)

uberfire-extensions/uberfire-layout-editor/uberfire-layout-editor-client/src/main/java/org/uberfire/ext/layout/editor/client/components/rows/Row.java
```diff
@@ -855,6 +866,61 @@ private void setupColumnResizeActions() {
...
+    public void clearProperties() {
+        properties.clear();
+        setupCssProperties();
+    }
+
+    @Override
+    public List<PropertyEditorCategory> getPropertyCategories() {
+        List<PropertyEditorCategory> result = new ArrayList<>();
+
+        PropertyEditorCategory category = layoutCssHelper.createCategory(LayoutEditorCssHelper.CSS_CATEGORY_PANEL);
```
ederign: Could you please extract this and the container getPropertyCategories() block to a helper class?

uberfire-extensions/uberfire-layout-editor/uberfire-layout-editor-client/src/main/java/org/uberfire/ext/layout/editor/client/infra/DragHelperComponentColumn.java
```diff
@@ -55,11 +63,38 @@
...
+                    dragHelper.lookupDragTypeBean(layoutComponent.getDragTypeName());
         }
         return layoutDragComponent;
     }
 
+    public List<PropertyEditorCategory> getLayoutDragComponentProperties() {
+        List<PropertyEditorCategory> result = getLayoutDragComponent().getPropertyCategories(layoutComponent);
+        Map<String,String> propertyMap = layoutComponent.getProperties();
+
+        PropertyEditorCategory category = layoutCssHelper.createCategory(LayoutEditorCssHelper.CSS_CATEGORY_PANEL);
```
ederign: please move also this!

uberfire-extensions/uberfire-layout-editor/uberfire-layout-editor-client/src/main/java/org/uberfire/ext/layout/editor/client/infra/LayoutEditorCssHelper.java
```diff
@@ -0,0 +1,235 @@
...
+
+        return propertyMap.entrySet().stream()
+            .filter(entry -> CssProperty.get(entry.getKey()) != null)
+            .map(entry -> new CssValue(entry.getKey(), entry.getValue()))
+            .collect(Collectors.toList());
+    }
+
+    // Validator classes
+
+    public PropertyFieldValidator LENGTH_VALIDATOR = new PropertyFieldValidator() {
```
ederign: extract to another class?

ederign:
@dgutierr great work! Just some minor comments and also could you please add some sample layouts (in txt) contains css definitions?

uberfire-extensions/uberfire-layout-editor/uberfire-layout-editor-client/src/main/java/org/uberfire/ext/layout/editor/client/infra/LayoutEditorCssHelper.java
dgutierr: Done => https://github.com/dgutierr/appformer/blob/926a6458c55e548441d99f2a311c3415a2234f43/uberfire-extensions/uberfire-layout-editor/uberfire-layout-editor-client/src/main/java/org/uberfire/ext/layout/editor/client/validator/LayoutEditorPropertyLengthValidator.java

dgutierr:
@kkufova  Option 2 done! :-)

dgutierr:
@ederign Fixed everything as suggested.

dgutierr:
@kkufova Fixed the preview issue with the nav components. There was a missing div id. => https://github.com/dgutierr/appformer/blob/d7588d5ff0c2a695ada1a909a276bd1c0878147d/uberfire-extensions/uberfire-layout-editor/uberfire-layout-editor-client/src/main/java/org/uberfire/ext/layout/editor/client/generator/AbstractLayoutGenerator.java#L46

Added a test as well to avoid regressions => https://github.com/dgutierr/appformer/blob/d7588d5ff0c2a695ada1a909a276bd1c0878147d/uberfire-extensions/uberfire-layout-editor/uberfire-layout-editor-client/src/test/java/org/uberfire/ext/layout/editor/client/generator/LayoutGeneratorTest.java

kkufova:
@dgutierr, thank you! The issue is gone. :-) I've found one more thing (I promise!): on a fresh instance, trying to change a width of a target div component (used by a tree) is causing an error. You need to close the page and open it again to be able to change the width. Check this:

https://issues.jboss.org/secure/attachment/12431902/PR188-bug2.webm

dgutierr:
@kkufova I verified that issue happens with brand new pages regardless the components used. I managed to reproduce it with two charts within the same row.

Can you please file a JIRA so that it is handled apart from this PR?

kkufova:
@dgutierr, I checked two kie-wb snapshots after this PR was merged and I cannot reproduce the issue anymore, hence I decided not to report it. Let's keep an eye on this though. :-)

PR #189, opened 2018-02-15
DROOLS-2323: Provide method to add form attribute with help

jomarko:
The issue [DROOLS-2323](https://issues.jboss.org/browse/DROOLS-2323)

@manstis 
This is ensemble:
- #189
- kiegroup/drools-wb#810


manstis:
@ederign Could you please approve and merge - I don't think it appropriate to refactor the ```FormStylePopup``` and related classes to Elemental2 so accept the use of GWT Widgets.

PR #193, opened 2018-02-20
DROOLS-2336 : [Project Oriented] Add REST tests back into kie-wb-distributions

Rikkola:
https://issues.jboss.org/browse/DROOLS-2336

https://github.com/kiegroup/kie-docs/pull/690
https://github.com/kiegroup/appformer/pull/193
https://github.com/kiegroup/kie-wb-distributions/pull/693


uberfire-rest/uberfire-rest-backend/src/main/java/org/guvnor/rest/backend/JobRequestScheduler.java
```diff
@@ -174,7 +171,7 @@ public void deployProjectRequest(final DeployProjectRequest jobRequest) {
                                          params));
     }
 
-    public void createOrganizationalUnitRequest(final SpaceRequest jobRequest) {
+    public void createSpaceRequest(final SpaceRequest jobRequest) {
```
jakubschwan: Is in this method used right operation? In this method is call command to create Org unit.

uberfire-rest/uberfire-rest-backend/src/main/java/org/guvnor/rest/backend/ProjectResource.java
```diff
@@ -422,91 +449,16 @@ public Response createSpace(Space space) {
...
-
-        return createAcceptedStatusResponse(jobRequest);
-    }
-
-    @POST
-    @Consumes(MediaType.APPLICATION_JSON)
-    @Produces(MediaType.APPLICATION_JSON)
-    @Path("/spaces/{spaceName}/")
-    @RolesAllowed({REST_ROLE, REST_PROJECT_ROLE})
-    public Response updateOrganizationalUnit(@PathParam("spaceName") String orgUnitName,
```
jakubschwan: There should be REST endpoint for updating of space. This REST endpoint is mention in restApi.jsp file in PR into kie-wb-distribtutions

uberfire-rest/uberfire-rest-backend/src/main/java/org/guvnor/rest/backend/JobRequestHelper.java
```diff
@@ -462,22 +451,21 @@ public JobResult removeOrganizationalUnit(final String jobId,
     }
 
     public JobResult createOrganizationalUnit(final String jobId,
-                                              final String organizationalUnitName,
-                                              final String organizationalUnitOwner,
-                                              final String defaultGroupId,
-                                              final List<String> repositoryNameList) {
+                                              final String spaceName,
```
jakubschwan: I am not sure if the name of the method is still actual. Shouldn't it be called createSpace?

uberfire-rest/uberfire-rest-backend/src/main/java/org/guvnor/rest/backend/JobRequestHelper.java
```diff
@@ -437,22 +426,22 @@ public JobResult deployProject(final String jobId,
     }
 
     public JobResult removeOrganizationalUnit(final String jobId,
-                                              final String organizationalUnitName) {
+                                              final String spaceName) {
```
jakubschwan: I'm not sure if the actual name of the method is correct. Shouldn't this method be called removeSpace?

uberfire-rest/uberfire-rest-backend/src/main/java/org/guvnor/rest/backend/JobRequestScheduler.java
Rikkola: The OrganizationalUnit is getting deleted and at the moment it is basically a Space.

uberfire-rest/uberfire-rest-backend/src/main/java/org/guvnor/rest/backend/ProjectResource.java
Rikkola: I need to double check if we need to remove that. Since the Space does not offer that much that can be edited.

uberfire-rest/uberfire-rest-backend/src/main/java/org/guvnor/rest/backend/JobRequestHelper.java
Rikkola: Maybe. I think it is better to rename it. Doing it tomorrow.

uberfire-rest/uberfire-rest-backend/src/main/java/org/guvnor/rest/backend/JobRequestHelper.java
Rikkola: I think it should, yes. Renamed.

uberfire-rest/uberfire-rest-backend/src/main/java/org/guvnor/rest/backend/JobRequestHelper.java
```diff
@@ -436,23 +425,23 @@ public JobResult deployProject(final String jobId,
...
-                                              final String organizationalUnitName) {
+    public JobResult removeSpace(final String jobId,
+                                 final String spaceName) {
         JobResult result = new JobResult();
         result.setJobId(jobId);
 
-        if (organizationalUnitName == null) {
+        if (spaceName == null) {
             result.setStatus(JobStatus.BAD_REQUEST);
             result.setResult("OrganizationalUnit name must be provided");
```
barboras7: I think that it would be better if the result message used "Space" instead of "OrganizationalUnit". What do you think?

uberfire-rest/uberfire-rest-backend/src/main/java/org/guvnor/rest/backend/JobRequestHelper.java
```diff
@@ -490,170 +478,67 @@ public JobResult createOrganizationalUnit(final String jobId,
             }
         }
 
-        OrganizationalUnit organizationalUnit = organizationalUnitService.getOrganizationalUnit(organizationalUnitName);
+        OrganizationalUnit organizationalUnit = organizationalUnitService.getOrganizationalUnit(spaceName);
         if (organizationalUnit != null) {
             result.setStatus(JobStatus.BAD_REQUEST);
-            result.setResult("OrganizationalUnit with name " + organizationalUnitName + " already exists");
+            result.setResult("OrganizationalUnit with name " + spaceName + " already exists");
```
barboras7: Same as my previous comment. Maybe it would be better to use "Space" in the result.

uberfire-rest/uberfire-rest-backend/src/main/java/org/guvnor/rest/backend/cmd/CreateOrgUnitCmd.java
```diff
@@ -41,14 +41,13 @@ public JobResult internalExecute(JobRequest request) throws Exception {
...
-                                                     jobRequest.getDefaultGroupId(),
-                                                     jobRequest.getProjects());
+            result = helper.createSpace(jobRequest.getJobId(),
+                                        jobRequest.getSpaceName(),
+                                        jobRequest.getOwner(),
+                                        jobRequest.getDefaultGroupId());
         } finally {
             JobStatus status = result != null ? result.getStatus() : JobStatus.SERVER_ERROR;
-            logger.debug("-----createOrganizationalUnit--- , OrganizationalUnit name: {}, OrganizationalUnit owner: {} [{}]",
+            logger.debug("-----createSpace--- , OrganizationalUnit name: {}, OrganizationalUnit owner: {} [{}]",
```
barboras7: Probably same as above. The log message would be more explanatory if "Space name" and "Space owner" were used.

uberfire-rest/uberfire-rest-backend/src/main/java/org/guvnor/rest/backend/cmd/RemoveOrgUnitCmd.java
```diff
@@ -41,11 +41,11 @@ public JobResult internalExecute(JobRequest request) throws Exception {
...
         JobResult result = null;
         try {
-            result = helper.removeOrganizationalUnit(jobRequest.getJobId(),
-                                                     jobRequest.getSpaceName());
+            result = helper.removeSpace(jobRequest.getJobId(),
+                                        jobRequest.getSpaceName());
         } finally {
             JobStatus status = result != null ? result.getStatus() : JobStatus.SERVER_ERROR;
-            logger.debug("-----removeOrganizationalUnit--- , OrganizationalUnit name: {}",
+            logger.debug("-----removeSpace--- , OrganizationalUnit name: {}",
```
barboras7: Same as already mentioned.

uberfire-rest/uberfire-rest-backend/src/main/java/org/guvnor/rest/backend/ProjectResource.java
```diff
@@ -531,7 +485,13 @@ public Response deleteSpace(@PathParam("spaceName") String spaceName) {
         return createAcceptedStatusResponse(jobRequest);
     }
 
-    private Response createAcceptedStatusResponse(JobRequest jobRequest) {
+    private void assertObjectExists(final Object o) {
+        if (o == null) {
+            throw new WebApplicationException(Response.status(Status.NOT_FOUND).entity(o).build());
```
jhrcek: Just wondering what's the use of adding `entity(o)` if it's always null. Wouldn't it be better to provide at least some string description so the "Not found" response is more useful?

jhrcek: @Rikkola ping in case you missed my question in the flood of comments :-)

Rikkola: Good point. Like a professional I just copy pasted this line in here. I'll take that out.

jhrcek: @Rikkola Wouldn't it make more sense to add some info about what was null? How about adding String parameter to the `assertObjectExists(final Object o, String whatIsMissing)` and adding this to the exception somehow? Especially since some methods can fail with multiple "missing objects", like `deleteProject` project can fail because space is missing or project within the space is missing - informing user which was the case would be useful!

Rikkola: Makes sense, I'll do this.

ederign:
@jhrcek could you please take a second look on this?

jhrcek:
@ederign My only objection to this PR was that in several places a method `assertObjectExists` is used, which leads to "anonymous" WebApplicationException being thrown which loses the information about what was missing.  I think we agreed with Toni @Rikkola that he'd add some kind of "What was missing" information to that exception.

Rikkola:
@jhrcek It actually is implemented, but I see how it is hard to see it here since it looks like I have forgotten to push the latest. 

Just did so, should be better now.

PR #196, opened 2018-02-22
AF-1036: implement sanitized number input

michpetrov:
New component - input that allows only numbers (and optionally minus sign and/or decimal point) to be typed. 

Issue: [AF-1036](https://issues.jboss.org/browse/AF-1036)

Two things to note:
 - `<input type="number">` is already sanitized by default in Google Chrome (but Chrome also allows you to type in "e" and a plus sign)
 - in Firefox it is possible to enter an invalid character using a compose key (e.g. Compose + o + a => å), even cancelling all keyboard events does not prevent this.

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/SanitizedNumberInput.java
+++ b/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/SanitizedNumberInput.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.uberfire.client.views.pfly.widgets;
+
+import javax.enterprise.context.Dependent;
+import javax.inject.Inject;
+
+import org.jboss.errai.common.client.api.IsElement;
+import org.jboss.errai.common.client.dom.EventListener;
+import org.jboss.errai.common.client.dom.KeyboardEvent;
+import org.jboss.errai.common.client.dom.NumberInput;
+
+@Dependent
+public class SanitizedNumberInput implements IsElement {
+
+    @Inject
+    NumberInput input;
+
+    private boolean allowNegative = false;
+    private boolean allowDecimal = false;
+
+    public void init() {
+        init("0", null, false, false);
+    }
+
+    public void init(final String min, final String step) {
+        init(min, step, false, false);
+    }
+
+    public void init(final String min, final String step, final boolean allowNegative, final boolean allowDecimal) {
+        if (min != null) {
+            input.setAttribute("min", min);
+        }
+        if (step != null) {
+            input.setAttribute("step", step);
+        }
+        this.allowDecimal = allowDecimal || (step != null && step.contains("."));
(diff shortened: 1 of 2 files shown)


uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/SanitizedNumberInput.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public void init(final String min, final String step, final boolean allowNegative, final boolean allowDecimal) {
+        if (min != null) {
+            input.setAttribute("min", min);
+        }
+        if (step != null) {
+            input.setAttribute("step", step);
+        }
+        this.allowDecimal = allowDecimal || (step != null && step.contains("."));
```
pefernan: @michpetrov the decimal separator depend on the local, maybe you could get it from LocaleInfo.getCurrentLocale().getNumberConstants().decimalSeparator();

Other than that I like the change.

michpetrov: @pefernan unless I'm doing something wrong `<input type="number">` is by itself not localized, i.e. I cannot use "," as a decimal separator even though that's valid in my locale (Chrome doesn't even let me type it). The parse functions in Java also don't take locale into account.

I could add the localization (at least for the decimal separator) but I feel that adds a lot of problems:
 - we'd have to use `<input type="text">` losing the built-in browser validation
 - I myself prefer to use "." even if that's not correct so I wouldn't want to force users to use their locale but if we allow more than one separator we wouldn't be able to parse the value correctly

pefernan: @michpetrov I can try it locally here in Barcelona we use the "," separator, is there any place where is it used on the workbench?

michpetrov: I was testing it in plain HTML but AFAIK none of the number inputs used in workbench have localization.

michpetrov:
@jomarko I'm not familiar with unit tests for GWT, [here](https://github.com/kiegroup/jbpm-wb/blob/457c793724ecd0f066cffd12b79912f08f101c8d/jbpm-wb-common/jbpm-wb-common-client/src/test/java/org/jbpm/workbench/common/client/filters/basic/BasicFiltersViewImplTest.java#L85-L115) is the test for the current implementation, is something like that sufficient?

jomarko:
@michpetrov yes, something like that is definitely sufficient, let me know when done I will re-review. Thanks.

michpetrov:
@jomarko I've added the test.

kiereleaseuser2:
Can one of the admins verify this PR? Comment with 'ok to test' to start the build.

kiereleaseuser2:
Can one of the admins verify this PR? Comment with 'ok to test' to start the build.

uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/SanitizedNumberInput.java
cristianonicolai: @pefernan you can check the input number behavior using any html page where an input number is available. It seems to me that the result will be the same, Chrome does not support using comma so @michpetrov suggestion seems reasonable. See also the known issues reported in here: https://caniuse.com/#feat=input-number

kiereleaseuser3:
Can one of the admins verify this PR? Comment with 'ok to test' to start the build.

kiereleaseuser3:
Can one of the admins verify this PR? Comment with 'ok to test' to start the build.

kiereleaseuser2:
Can one of the admins verify this PR? Comment with 'ok to test' to start the build.

kiereleaseuser2:
Can one of the admins verify this PR? Comment with 'ok to test' to start the build.

jsoltes:
@ederign do you know what is this pr waiting for?

ederign:
@jsoltes this was not approved by anyone from engineering. /cc @cristianonicolai @pefernan

PR #198, opened 2018-02-22
JBPM-6815: Stunner - Ability to import/export process

wmedvede:


wmedvede:
Hi @romartin @hasys would you mind review this PR?
Thanks.

wmedvede:
@hasys an additional parameter was needed to pass to the FileSaver.js component that is being using for performing the client side downloading.

hasys:
@wmedvede I see..

So, am I correct that now everyone who used Uberfire download feature before will lose their [Byte order mark](https://en.wikipedia.org/wiki/Byte_order_mark) everywhere without any notification?

wmedvede:
@hasys well I'd say that in general we don't want to download files with additional extra characters that are not part of the original file. I believe this apply for all situations where we download files with this component. And this is more or less the only available parameter we have to pass to this component to avoid something that as is was shown causes errors. Other than this I don't know what more to say.

hasys:
@wmedvede yes, just curious why no one before realize this strange additional extra character. Hope no one rely on it and this change won't break something to someone. From ours side we can make this invocation with default value, and call this method with non default value from our's code.

But in general I am agree with you. Let's wait for the Appformer maintainers vote.

wmedvede:
see that it also says "The Unicode Standard permits the BOM in UTF-8,[3] but does not require or recommend its use"

wmedvede:
I closed it by mistake, open again

PR #208, opened 2018-02-28
DASHBUILDE-264 - Allow NUMBER columns to be used in displayers

mswiderski:
@dgutierr here are the changes we talked about, hopefully that is all that is needed.


dgutierr:
@mswiderski I verified the changes work well. Notice though, as it is implemented right now the grouping by number columns is doing nothing but treating every numeric value in the dataset as a category, same as text columns where each distinct string is a category.  So be careful with numeric columns that have a large amount of different values as this might cause performance issues while rendering the chart.

dgutierr:
@kkufova Can you take a look please?

mswiderski:
thanks @dgutierr and that's fine - at least for the use case as see it's valid - process instance state (4 values) sla compliance (4 values) and similar.

dgutierr:
@ederign Can you please merge? I don't have permissions :-(

PR #215, opened 2018-03-01
Remove duplicated junit/mockito/fest dependency declarations

jhrcek:
@ederign can you please review and merge? This removes all the duplicate declarations of test dependencies junit / mockito-core / fest-assert-core.
All descendants of uberfire-parent automatically inherit these thanks to [this declaration](https://github.com/kiegroup/appformer/blob/master/pom.xml#L1335-L1351) in uberfire-parent, so no need to declare those in descendant poms.

After this PR is merged I'm planning to globally replace all usages of fest-assert-core (which is predecessor of assertj, and is no longer maintained since 2013) by assertj-core.


jhrcek:
jenkins retest this

PR #218, opened 2018-03-01
AF-985: [Project Oriented] Project doesn't show all imported assets.

mbarkley:


mbarkley:
Related to https://github.com/kiegroup/kie-wb-common/pull/1442

mbarkley:
This PR fixes batch indexing, which was not working as intended (details below) which lead to issues loading assets in the library project screen immediately after a project was first imported.

First this PR removes the `BatchIndex` class. Because the batch index was scheduled through an executor service, it would often run before any files had been added into a file system.

Instead, the file-system watch service now uses batch indexing in the MetadataIndexEngine. When files are imported to a new repository, typically they are copied using the batch setting on the IOService, so the watch service should see all the files at the same time, in effect creating a batch indexing when a repo is first created.

I also made changes to how batches work in the MetadataIndexEngine so that a batch is coupled to a single cluster and thread. When a batch is committed in the MetadataIndexEngine, a CDI event is fired that is observed by the library backend in the linked kie-wb-common PR.

@porcelli @ederign @adrielparedes Would any of you care to review this?

PR #219, opened 2018-03-02
Replace fest-assert-core with assertj-core

jhrcek:
@ederign please check. This is the followup to fest-assert cleanup as promised yesterday.

- got rid of fest-assert-core completely
- replaced all occurrences with assertJ
- removed duplicate declarations of assertj-core from poms (we only need one in uberfire parent)
- cleaned up assertJ usage (e.g better way to test for exceptions being thrown, removed redundant `.isNotNull()` calls, 
```
assertThat(dir.toFile().exists()).isTrue(); -> assertThat(dir.toFile()).exists();
```
etc.
Note that sometimes it was necessary to use `AssertionsForClassTypes` instead of `Assertions`, because e.g. `Path` is also `Iterable` so `Assertions.assertThat(path)...` was ambiguous

### DIFF
--- a/uberfire-nio2-backport/uberfire-nio2-api/src/test/java/org/uberfire/java/nio/file/FilesTest.java
+++ b/uberfire-nio2-backport/uberfire-nio2-api/src/test/java/org/uberfire/java/nio/file/FilesTest.java
@@ -62,31 +62,40 @@ public class FilesTest extends AbstractBaseTest {
         }
     }
 
-    @Test(expected = NoSuchFileException.class)
+    @Test
     public void newInputStreamNonExistent() {
-        Files.newInputStream(Paths.get("/path/to/some/file.txt"));
+        assertThatThrownBy(() -> Files.newInputStream(Paths.get("/path/to/some/file.txt")))
+                .isInstanceOf(NoSuchFileException.class);
     }
 
-    @Test(expected = NoSuchFileException.class)
+    @Test
     public void newInputStreamOnDir() {
         final Path dir = newTempDir();
-        Files.newInputStream(dir);
+        assertThatThrownBy(() -> Files.newInputStream(dir))
+                .isInstanceOf(NoSuchFileException.class);
     }
 
-    @Test(expected = IllegalArgumentException.class)
+    @Test
     public void newInputStreamNull() {
-        Files.newInputStream(null);
+        assertThatThrownBy(() -> Files.newInputStream(null))
+                .isInstanceOf(IllegalArgumentException.class)
+                .hasMessage("Parameter named 'path' should be not null!");
     }
 
-    @Test(expected = org.uberfire.java.nio.IOException.class)
+    @Test
     public void newOutputStreamOnExistent() {
         final Path dir = newTempDir();
-        Files.newOutputStream(dir);
+
+        assertThatThrownBy(() -> Files.newOutputStream(dir))
+                .isInstanceOf(org.uberfire.java.nio.IOException.class)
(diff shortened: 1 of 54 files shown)


jhrcek:
jenkins retest this

jhrcek:
Rebased on master. @ederign can you please check this PR when you have a moment?

jhrcek:
Rebased on master again. @mbarkley or @porcelli can you please check and merge this? I'm unable to catch @ederign

uberfire-nio2-backport/uberfire-nio2-api/src/test/java/org/uberfire/java/nio/file/FilesTest.java
```diff
@@ -138,14 +138,13 @@ public void createDirectory() {
         final Path dir = Files.createDirectory(path.resolve("myNewDir"));
 
         assertThat(dir).isNotNull();
-        assertThat(dir.toFile().exists()).isTrue();
-        assertThat(dir.toFile().isDirectory()).isTrue();
+        assertThat(dir.toFile()).exists();
```
porcelli: @jhrcek by just invoing the methos... will `true` be checked?

uberfire-nio2-backport/uberfire-nio2-api/src/test/java/org/uberfire/java/nio/file/FilesTest.java
```diff
@@ -528,55 +513,42 @@ public void moveFileInvalidSourceAndTarget() throws IOException {
...
     }
 
     @Test(expected = IllegalArgumentException.class)
-    public void moveNull2() throws IOException {
+    public void moveNull2() {
         Files.move(null,
                    newTempDir());
     }
 
     @Test(expected = IllegalArgumentException.class)
```
porcelli: why preserving here the "expected"? left over?

uberfire-nio2-backport/uberfire-nio2-api/src/test/java/org/uberfire/java/nio/file/FilesTest.java
jhrcek: @porcelli before the change it was calling `exists` method of the `Path` class. Now it's calling `exists` of AbstractPathAssert [javadoc](http://joel-costigliola.github.io/assertj/core/api/org/assertj/core/api/AbstractPathAssert.html#exists()) which gives better error messages in case the assertion fails. Previously it would give "expected true, but was false", now if it fails it would give something line "Expecting file [specific path] to exist".

uberfire-nio2-backport/uberfire-nio2-api/src/test/java/org/uberfire/java/nio/file/FilesTest.java
jhrcek: Well spotted, I must have missed that. I'll replace the remaining occurrences tomorrow.

triceo:
@jhrcek I can set reviewers. Unfortunately this particular permission, I think, is only available to project administrators or some other high-ranking role within the project - and therefore it is not widely distributed. :-(

jhrcek:
@porcelli I replaced more than a hundred occurrences of `@Test(expected ...)` by `Assertions.assertThatThrownBy(...)`. But there are few hundred occurrences left in the entire repo. Since the goal of this PR is primarily to get rid of the outdated dependency (fest-assert-core), and not to refactor all the tests to use assertj, I consider this to be ready for merge.

PR #222, opened 2018-03-02
Fixes typos

mareknovotny:


uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-security-management-keycloak/README.md
```diff
@@ -5,7 +5,7 @@ Introduction
 ------------
 The classes in this package contain the user and group manager provider implementation for JBoss KeyCloak.              
 
-There exist two implementation for this provider:           
+There are existing two implementation for this provider:
```
jhrcek: I believe the correct would be "There are two implementation of this provider"

mareknovotny: yep, that's even better ;)

PR #226, opened 2018-03-07
AF-1072: Log the file system path in JGitFileSystemProvider

jesuino:
AF-1072: Log the file system path in JGitFileSystemProvider


ederign:
please jenkins retest this

PR #234, opened 2018-03-12
JBPM-7002 - Group action buttons in list table to kebab

cristianonicolai:
- Add drop position and separator helpers to KebabMenu
- Change button element to be used as ManagedInstance
- Update tests

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/Elemental2Producer.java
+++ b/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/Elemental2Producer.java
@@ -1,0 +1,16 @@
+package org.uberfire.client.views.pfly.widgets;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.inject.Produces;
+
+import elemental2.dom.DomGlobal;
+import elemental2.dom.HTMLDocument;
+
+@ApplicationScoped
+public class Elemental2Producer {
+
+    @Produces
+    public HTMLDocument produceDocument() {
+        return DomGlobal.document;
+    }
+}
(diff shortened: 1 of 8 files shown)


cristianonicolai:
Please wait until https://github.com/cristianonicolai/jbpm-wb/pull/33 is pushed into master to merge.

uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/Elemental2Producer.java
```diff
@@ -0,0 +1,16 @@
...
+import javax.enterprise.inject.Produces;
+
+import elemental2.dom.DomGlobal;
+import elemental2.dom.HTMLDocument;
+
+@ApplicationScoped
+public class Elemental2Producer {
+
+    @Produces
+    public HTMLDocument produceDocument() {
```
jomarko: Just question, do we need this method to make code testable? In other words, I am asking if for example `(HTMLLIElement) document.createElement("li");` can't be achieved if that class has code:
```java
@Inject
private Instance<HTMLLIElement> liElements;

// ...

liElements.get();
```

cristianonicolai: correct the goal is to improve the test code, although my understanding is that using ManagedInstance or Instance usually only works for custom components. When using it for elemental2 elements (HTMLLIElement, span, etc...) I usually run into _IOCResolutionException: Multiple beans matched com.google.gwt.core.client.JavaScriptObject$_, that's why I continue to use _document.createElement_ in this case. Maybe @mbarkley  could help describe if this is the best approach.

mbarkley: @jomarko Unfortunately `Instance<T>` doesn't work for any class `T` that is a native JS type (i.e. `@JsType(isNative = true)`) which includes all element wrappers in Elemental 2 and Errai. At present there isn't a better way to do this than injecting and using the document object.

jomarko: Great, thanks for explanation.

PR #241, opened 2018-03-14
JBPM-6876: Stunner - easy way to migrate from Stunner to current Designer

wmedvede:
- Adds the ability to optionally show an inline notification in the ConfirmPopup according to UX requeriment.

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/Button.java
+++ b/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/Button.java
@@ -16,6 +16,8 @@
 
 package org.uberfire.client.views.pfly.widgets;
 
+import java.util.stream.Stream;
+import javax.annotation.PostConstruct;
 import javax.enterprise.context.Dependent;
 import javax.inject.Inject;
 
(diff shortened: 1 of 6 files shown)


wmedvede:
Hello @romartin @hasys would you mind review this PR?
Thanks.

uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/Button.java
```diff
@@ -50,6 +59,8 @@ public void setType(final ButtonType type) {
     }
 
     public void setButtonStyleType(final ButtonStyleType type) {
+        removeClass("btn");
```
jomarko: Just question, why we remove this "btn" class when changing button style, but we declare this class by default, for example in `ConfirmPopup.html`:
```html
<button type="button" class="btn btn-default" data-field="confirm-cancel" data-dismiss="modal"></button>
<button type="button" class="btn btn-danger" data-field="confirm-ok"></button>
```

uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/ConfirmPopup.java
```diff
@@ -36,27 +39,70 @@
...
+    @PostConstruct
+    public void init() {
+        cancelButton.setText(translationService.getTranslation(Constants.ConfirmPopup_Cancel));
+    }
+
     public void show(String title,
                      String okButtonText,
                      String confirmMessage,
                      final Command okCommand) {
+        show(title,
```
jomarko: Minor thing, could we please use `final` keyword for arguments of both `show` methods. I think it is good to be consistent with style of declaration `@EventHandler` methods in this class.

uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/widgets/Button.java
wmedvede: The "btn" class is also removed for safety and to be sure that all the elements that configure the button style are properly set to the new style.

PR #243, opened 2018-03-15
[RHBA-523]: HEAD branch is not copied anymore when synchronizing remote branches

adrielparedes:
When a repository is cloned it has some local branches and some remote branches.

This is an example:
```
* master                827d3cf 6.4.x compatible DataObjects for testing RHBPMS-4837
  remotes/origin/HEAD   -> origin/master
  remotes/origin/master 827d3cf 6.4.x compatible DataObjects for testing RHBPMS-4837
```

The problem was the HEAD remote branch. That is not a real branch, it is just a symbolic reference to origin/master. This could be changed or added again, but is not completely necessary. 
So when it tried to create branches two of them were pointing to the same tree node, so the result was "NO_CHANGE" and it couldn't create the new branch. Process failed after this.

@mbarkley Would you mind to review it? Thanks!


jhrcek:
@ederign can you please check and merge?

adrielparedes:
@jhrcek and @tomasdavidorg confirmed that the issue is not reproducible anymore. Closing PR.

PR #244, opened 2018-03-15
RHBA-506: Use Initial context to retrieve connection factory

sutaakar:
@ederign Please take a look on this.

There are two possibilities how connect into the JMS broker.

- Use local Initial context. This is the default option, use Initial context of application server where the Workbench is running. External JMS broker can be specified using this option by adding connection factory backed by proper resource adapter to the application server.

- Use remote Initial context. This can be used for the case if the user wants to connect to external application server. In this case the user needs to add proper connection factory classes into server classpath (for example adding `org.apache.activemq.artemis` module as dependency of the Workbench).

By default there is used connection factory with JNDI `java:/ConnectionFactory` from local context, which is available by default.

### DIFF
--- a/uberfire-commons/pom.xml
+++ b/uberfire-commons/pom.xml
@@ -50,56 +50,14 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
-    <!-- ActiveMQ (artemis) -->
-    <dependency>
-      <groupId>org.apache.activemq</groupId>
-      <artifactId>artemis-jms-client</artifactId>
-      <exclusions>
-        <exclusion>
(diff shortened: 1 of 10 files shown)


spolti:
The changes looks nice to me.
It will help us to made the cluster setup on OpenShift. In the current way the class was wrote not work for a external broker that needs to use a resource adapter, with this changes this configuration will be more flexible.

porcelli:
@sutaakar did you manage to try the MaaS?

sutaakar:
@porcelli Not yet, I will test that today or in the Monday.

sutaakar:
Replaced by https://github.com/kiegroup/appformer/pull/325

PR #245, opened 2018-03-16
JBPM-6886: Update Dashboards layout. toolbar icons style

nmirasch:
Updated styles for multiscreen toolbar icons


nmirasch:
@cristianonicolai  could you take a look? Thanks!

PR #246, opened 2018-03-16
[AF-1102] Git repos disappearing when you clone a repo twice with a invalid remote in the second try

ederign:


jhrcek:

porcelli:
good catch @ederign and @jhrcek - this issue happened since ever, thx for spot this issue.

ederign:
@jhrcek thanks for the hint and the detailed reproducer.

PR #249, opened 2018-03-21
Fix missing button text and icon

cristianonicolai:


cristianonicolai:
@wmedvede reverting this part of the change because it overrides the text that is set when using this element in an html template. Would it work for you?

@nmirasch @bsremac fyi

wmedvede:
Hi @cristianonicolai , could you keep this code instead? 
I think it'l fit your needs and also we need to have the ability to change the button text dinamically.



   //let the text node be a class field. 
    private Text textNode;

    public void setText(final String text) {
        if (textNode == null) {
           //create it only the first time setText is invoked
            textNode = document.createTextNode(text);
            getElement().appendChild(textNode);
        } else {
           //other times just set the new text.
            textNode.data = text;
        }
    }

Thanks!

cristianonicolai:
@wmedvede makes sense to me, updated accordingly. Thanks

ederign:
@jhrcek @cristianonicolai  who is the QE that I should assign jbpm related PR's?

cristianonicolai:
@ederign if it is jBPM UI related that would be @bsremac

ederign:
@bsremac could you please take a look at this?

PR #262, opened 2018-03-27
AF-1116: use a proper ssh default algorithm

porcelli:


jhrcek:
@tomasdavidorg this should make the setting of this property in our automation superfluous. Please remove it and verify this new default works.

tomasdavidorg:
Thanks @porcelli

PR #266, opened 2018-03-28
JBPM-7103 - KIE Server custom queries exposed as DataSet editor in wo…

mswiderski:
…rkbench

@dgutierr could you please have a look?


ederign:
@dgutierr can you please take a look on this?

PR #258, opened 2018-03-23
RHDM-390,RHDM-391: Workaround for ERRAI-1101

yurloc:
Related PRs:
kiegroup/appformer/pull/258
kiegroup/kie-wb-common/pull/1535
kiegroup/drools-wb/pull/834
kiegroup/optaplanner-wb/pull/266
kiegroup/jbpm-wb/pull/1011
kiegroup/kie-wb-distributions/pull/716

Works around: https://issues.jboss.org/browse/ERRAI-1101
Fixes: https://issues.jboss.org/browse/RHDM-390, https://issues.jboss.org/browse/RHDM-391

PR #267, opened 2018-03-28
RHDM-390,RHDM-391: Work around ERRAI-1101 using antrun 

yurloc:
Reverts and supersedes #258.

This alternative solution will have zero maintenance demand and shouldn't pose a risk of breaking Zanata integration.

Requesting approval from @mbarkley (Errai expert) and @manstis (Zanata expert).

Related PRs:
kiegroup/appformer/pull/267
kiegroup/droolsjbpm-build-bootstrap/pull/729
kiegroup/kie-wb-common/pull/1553
kiegroup/drools-wb/pull/837
kiegroup/jbpm-wb/pull/1026
kiegroup/optaplanner-wb/pull/268
kiegroup/kie-wb-distributions/pull/720


yurloc:
@manstis The copy is created during `process-resources` phase which comes _before_ `compile`. I don't know how the execution order of goals that are bound to the same phase is determined. So if Zanata pull is executed in `process-resources` too, there is a chance it comes after the workaround.

mbarkley:
IIRC, the are executed in the order declared if in the same phase, so as it is now your copying should run after Zanata.

yurloc:
@mbarkley Yes we can. I wanted to only declare it UI related repositories but probably it's not worth the duplication. Even more so that it's only a temporary workaround.

I'm going to open backport PRs for 7.7.x and 7.5.x so I'll do it for appformer and kie-parent POMs in the new PRs.

If you're OK with that, please approve this PR (kie-parent is not appformer's parent).

yurloc:
@mbarkley One question: is there any difference between declaring the plugin execution in `build/plugins` and `build/pluginManagement/plugins`?

mbarkley:
IIRC, when you declare it in pluginManagement, the plugin inherits the
configuration *if* it is used. You then still need to add the plugin in
build/plugins somewhere to use it (but without any configuration).

yurloc:
@mbarkley Yes that's it, I forgot about it. So in this case it should go into build/plugins cause we don't want this configuration to be inherited by child executions (should there be any declared in the future).

mbarkley:
Configuration in build/plugins is also inherited. The difference between it
and management, is that management only declares a configuration, whereas
build/plugins in a parent declares a configuration that will be executed.

If you want to only specify the configuration once, but control where it is
executed more carefully, you should define your execution in
pluginManagement but don't bind it to a phase.

Then in repositories where it should be used (kie-wb-common, etc.) you can
bind the execution (inherited from the parent) to a phase.

yurloc:
@mbarkley I've replaced all the other PRs by moving the plugin configuration to `kie-parent` in kiegroup/droolsjbpm-build-bootstrap#729. This PR should be merged as kie-parent is not a parent of AppFormer projects. Can you approve this?

ederign:
@manstis can you please recheck this?

PR #269, opened 2018-03-28
[AF-1087]: Added VisibleAssetQualifier to ResourceTypeDefinitions

adrielparedes:
Related to:
https://github.com/kiegroup/appformer/pull/269
https://github.com/kiegroup/kie-wb-common/pull/1550
https://github.com/kiegroup/drools-wb/pull/838
https://github.com/kiegroup/optaplanner-wb/pull/269
https://github.com/kiegroup/jbpm-designer/pull/760

### DIFF
--- a/uberfire-api/src/main/java/org/uberfire/workbench/annotations/VisibleAsset.java
+++ b/uberfire-api/src/main/java/org/uberfire/workbench/annotations/VisibleAsset.java
@@ -1,0 +1,25 @@
+package org.uberfire.workbench.annotations;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+import javax.inject.Qualifier;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * This annotation represents a Resource that should be visible to end users. Any other resource not annotated with it
+ * should not be shown. In case you need to inject a Resource that contains @VisibleAsset in an object that doesn't need
+ * this concept you can add @Default qualifier to it.
+ */
+@Documented
+@Qualifier
+@Retention(RUNTIME)
+@Target({TYPE, PARAMETER, FIELD, METHOD})
+public @interface VisibleAsset {
+
+}
--- a/uberfire-api/src/main/java/org/uberfire/workbench/type/TextResourceTypeDefinition.java
+++ b/uberfire-api/src/main/java/org/uberfire/workbench/type/TextResourceTypeDefinition.java
@@ -16,10 +16,15 @@
 
 package org.uberfire.workbench.type;
 
+import javax.enterprise.inject.Default;
+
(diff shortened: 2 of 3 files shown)


adrielparedes:
jenkins full downstream build

adrielparedes:

adrielparedes:
@mbarkley @jhrcek I've uploaded the PR and I created some other because of that change. Also I execute a downstream build that is failing but is not related to my changes. In the PR description is the full list with PRs.

Would you mind to review them?

uberfire-api/src/main/java/org/uberfire/workbench/annotations/VisibleAsset.java
```diff
@@ -0,0 +1,16 @@
...
+import java.lang.annotation.Target;
+import javax.inject.Qualifier;
+
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Qualifier
+@Retention(RUNTIME)
+@Target({TYPE, PARAMETER})
```
mbarkley: Targets for a qualifier should include `FIELD` and `METHOD`.

uberfire-api/src/main/java/org/uberfire/workbench/annotations/VisibleAsset.java
```diff
@@ -0,0 +1,16 @@
...
+public @interface VisibleAsset {
```
mbarkley: Could you please add `@Documented` and javadoc to this qualifier? Things worth mentioning in the javadoc:

- What is this qualifier used for? (Don't have to mention the library, but maybe more generally that this means that this type of resource is a file that should be visible to end users.)
- Why you should still add `@Default` when using this qualifier.

adrielparedes: @mbarkley thanks for the comments, I will fix them right now

porcelli:
@adrielparedes as @ederign mentioned, I'd like to understand why this way, and not the other way around? As sounds to me that most files we want to be visible, and just the minoroty we don't.
I'm wondering about a new file that we might support... like DMN graphic editor, do we want annotate it to be Visible? OR we by default we'd assume it will be visible unless you explict as to @Hide it.

This is not a request for change (yet), just want start a discussion.

@mbarkley @ederign @adrielparedes @tiagobento please all speak up :)

mbarkley:
I think the best argument in favour of the current approach is ease of implementation. You can inject for qualifiers that are there but not for ones that are not.

If we make visible definitions have the  default qualifier and others non-default then we have to change a bunch of other injection sites that use specific type definitions.

You could inject hidden/system definitions and check that something shouldn't be shown in the library indexer but that wouldn't be as robust as what we have now (POM files would be visible, for example, because we have no POM type).

So IMO this is the easiest approach to implement. If we want to do an approach that is opt-out instead of opt-in, I would suggest we forget qualifiers and add a method to the type definition interface (ex. isVisible) and filter based on it.

porcelli:
thank you @mbarkley for the clear articulation of the solution.

based on that... +1 :)

ederign:
@jhrcek could you please take a look at this?

jhrcek:
For some reason there are no workbench artifacts in full downstream build. Trying again.. 

PR #275, opened 2018-04-04
RHBA-415: After rename, designer is not marked as dirty more

karreiro:
See https://issues.jboss.org/browse/RHBA-415

Before:
before-1
before-2

After:
after-1
after-2

---

Part of an ensemble:
- https://github.com/kiegroup/appformer/pull/275
- https://github.com/kiegroup/drools-wb/pull/842

### DIFF
--- a/uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/test/java/org/uberfire/ext/editor/commons/client/BaseEditorTest.java
+++ b/uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/test/java/org/uberfire/ext/editor/commons/client/BaseEditorTest.java
@@ -20,6 +20,7 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.function.Supplier;
 
+import com.google.gwt.user.client.ui.IsWidget;
 import com.google.gwtmockito.GwtMockitoTestRunner;
 import org.jboss.errai.common.client.api.Caller;
 import org.junit.Test;
@@ -28,13 +29,19 @@ import org.mockito.InjectMocks;
 import org.mockito.Mock;
 import org.uberfire.backend.vfs.ObservablePath;
 import org.uberfire.backend.vfs.Path;
+import org.uberfire.client.callbacks.Callback;
+import org.uberfire.client.workbench.events.ChangeTitleWidgetEvent;
+import org.uberfire.client.workbench.type.ClientResourceType;
 import org.uberfire.ext.editor.commons.client.history.VersionRecordManager;
 import org.uberfire.ext.editor.commons.client.menu.BasicFileMenuBuilder;
 import org.uberfire.ext.editor.commons.client.menu.common.SaveAndRenameCommandBuilder;
 import org.uberfire.ext.editor.commons.client.validation.Validator;
 import org.uberfire.ext.editor.commons.file.DefaultMetadata;
+import org.uberfire.java.nio.base.version.VersionRecord;
+import org.uberfire.mocks.EventSourceMock;
 import org.uberfire.mvp.Command;
 import org.uberfire.mvp.ParameterizedCommand;
+import org.uberfire.mvp.PlaceRequest;
 import org.uberfire.workbench.model.menu.MenuItem;
 
 import static org.junit.Assert.assertEquals;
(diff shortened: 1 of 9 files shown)


karreiro:
@mantis Precisely, Mike, that's the issue. The lock was not being released, and the editor was never marked as locked again.

There are two essential scenarios to cover:

- 1) The GDT Editor scenario:
    - Create a Guided Decision Table;
    - Rename the table;
    - Append some rows and add value to them;
    - (!) The editor was not being locked here.

- 1) Any other editor scenario (let's pick the DRL Editor):
    - Create a DRL file with the name "file";
    - Rename the DRL to "file1";
    - Make some changes in the file and save it;
    - Rename the DRL to "file2";
    - (!) The editor was not being locked and the title was not changing from "file1.drl - DRL" to "file2.drl - DRL".

uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/test/java/org/uberfire/ext/editor/commons/client/BaseEditorTest.java
```diff
@@ -369,6 +381,134 @@ public void testMakeMenuBarWhenItDoesNotContainAllMenuItems() {
...
+        assertEquals(editorTitle, event.getTitleDecoration());
+    }
+
+    @Test
+    public void testInitVersionRecordManager() {
+
+        final ObservablePath path = mock(ObservablePath.class);
+        final PlaceRequest placeRequest = mock(PlaceRequest.class);
+        final String version = "version";
+        final Callback<VersionRecord> callback = (v) -> {
```
manstis: Might have been more readable to simply use a mock Callback?

uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/mvp/LockManagerImpl.java
```diff
@@ -273,21 +273,33 @@ private boolean isLockedByCurrentUser() {
         return lockInfo.isLocked() && lockInfo.lockedBy().equals(user.getIdentifier());
     }
 
-    private void updateLockInfo(@Observes LockInfo lockInfo) {
-        if (lockTarget != null && lockInfo.getFile().equals(lockTarget.getPath())) {
+    void updateLockInfo(final @Observes LockInfo lockInfo) {
+        if (getLockTarget() != null && lockInfo.getFile().toURI().equals(lockTarget.getPath().toURI())) {
```
manstis: You make a point of using ```toURI()``` on the paths here; but.....

uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/mvp/LockManagerImpl.java
```diff
@@ -309,6 +321,12 @@ void onSaveInProgress(@Observes SaveInProgressEvent evt) {
         }
     }
 
+    void onRenameInProgress(@Observes RenameInProgressEvent event) {
+        if (getLockTarget() != null && event.getPath().equals(lockTarget.getPath())) {
```
manstis: ... here you're OK using ```Path``` equality? (not URI, like your change above). Is this OK?

karreiro:
@jomarko Good point. I'm not sure either. IMHO, the versions should be tracked right after every IO operation and not be lost.

Since it's not a regression, I think that this is not a blocker to this PR (but maybe we could raise a JIRA for that), see:

Before:
2018-04-05 6 46 45 pm

After:
2018-04-05 6 07 54 pm

uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/test/java/org/uberfire/ext/editor/commons/client/BaseEditorTest.java
karreiro: Good catch. Fixed.

uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/mvp/LockManagerImpl.java
karreiro: @manstis Here we can have a `PathImpl` being compared to an `ObservablePath` (that's what we're avoiding), and......

uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/mvp/LockManagerImpl.java
karreiro: ...here, we _always_ have an `ObservablePath` being compared to another `ObservablePath`.

I added a commentary to make this less confusing.

Thanks for the comment :-)

PR #276, opened 2018-04-04
[AF-1091]: Fixed error that appears after using the Contributors tab

adrielparedes:
The error happened when saving contributor in Contributors Tab. In that moment an UpdatedOrganizationalUnit event is thrown, and it contains the Organization Unit updated. 

WorkspaceProjectContext observes that event and update the context, but in that method all values are set to null but Organizational Unit.

I kept with the old active values but organizational unit because that is the only one that changed.

@barboras7 @Rikkola would you mind to review it please?

### DIFF
--- a/uberfire-project/uberfire-project-client/src/main/java/org/guvnor/common/services/project/client/context/WorkspaceProjectContext.java
+++ b/uberfire-project/uberfire-project-client/src/main/java/org/guvnor/common/services/project/client/context/WorkspaceProjectContext.java
@@ -67,11 +67,19 @@ public class WorkspaceProjectContext {
     }
 
     public void onOrganizationalUnitUpdated(@Observes final UpdatedOrganizationalUnitEvent event) {
-        contextChangeEvent.fire(new WorkspaceProjectContextChangeEvent(event.getOrganizationalUnit()));
+        WorkspaceProject updatedWorkspaceProject = new WorkspaceProject(event.getOrganizationalUnit(),
+                                                                        activeWorkspaceProject.getRepository(),
+                                                                        activeWorkspaceProject.getBranch(),
+                                                                        activeWorkspaceProject.getMainModule());
+        contextChangeEvent.fire(new WorkspaceProjectContextChangeEvent(updatedWorkspaceProject,
+                                                                       activeModule,
+                                                                       activePackage));
     }
 
     public void onProjectContextChanged(@Observes final WorkspaceProjectContextChangeEvent event) {
-        WorkspaceProjectContextChangeEvent previous = new WorkspaceProjectContextChangeEvent(activeWorkspaceProject, activeModule, activePackage);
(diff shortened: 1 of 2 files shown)


adrielparedes:

barboras7:

Rikkola:
Ok. This bug was made by me. Update should keep the old values ( almost ).

The `WorkbenchProject` also has the OU information. Here it is not updated. Easy fix might be to just create a new `WorkbenchProject` in the `WorkspaceProjectContext.java` in the `onOrganizationalUnitUpdated` method. Use the OU from the event and everything else from `activeWorkspaceProject`.

Rikkola:
And have test for it.

adrielparedes:
@Rikkola and why not to add that behavior in the method "onChange"? May be there checking if values are null or not I can override old values, wdyt?

adrielparedes:

adrielparedes:
@Rikkola I'm still working on something, do not review. Because is not enought to create a new WorkspaceProject and pass it to the WorkspaceProjectContextChangeEvent, because the onChange method uses getModule, and the module there is not from WorkspaceProject, tomorrow I will ping you again

Rikkola:
@adrielparedes No problem.

adrielparedes:

uberfire-project/uberfire-project-client/src/main/java/org/guvnor/common/services/project/client/context/WorkspaceProjectContext.java
```diff
@@ -67,19 +67,28 @@ public WorkspaceProjectContext(final Event<WorkspaceProjectContextChangeEvent> c
     }
 
     public void onOrganizationalUnitUpdated(@Observes final UpdatedOrganizationalUnitEvent event) {
-        contextChangeEvent.fire(new WorkspaceProjectContextChangeEvent(event.getOrganizationalUnit()));
+        WorkspaceProject updatedWorkspaceProject = new WorkspaceProject(event.getOrganizationalUnit(),
+                                                                        activeWorkspaceProject.getRepository(),
+                                                                        activeWorkspaceProject.getBranch(),
+                                                                        activeWorkspaceProject.getMainModule());
+        contextChangeEvent.fire(new WorkspaceProjectContextChangeEvent(updatedWorkspaceProject,
```
adrielparedes: @Rikkola I still have to add the other active modules and packages, but I took you advice and I've created the new WorkspaceProject

ederign:
@Rikkola can you please take a look on this?

adrielparedes:
@Rikkola no, I think it is done, thanks for the review :)

Rikkola:
jenkins please retest this

adrielparedes:

ederign:
@adrielparedes this will be ported for 7.7.x right?

adrielparedes:
@ederign right now

PR #288, opened 2018-04-11
[AF-1103]: Fixed issue that redirects to the list of assets when switching to project view

adrielparedes:
Added **requiresRefresh** attribute to know if changes in context needs to refresh to use the new data.


adrielparedes:

PR #285, opened 2018-04-09
AF-1120: Support standalone repositories as targets for the migration…

mbarkley:
… tool

PR #290, opened 2018-04-12
AF-1120: Support standalone repositories as targets for the migration…

mbarkley:
… tool (#285)


mbarkley:
Related: https://github.com/kiegroup/kie-wb-common/pull/1627

PR #292, opened 2018-04-12
[7.7.x] AF-1049: [Layout editor] Changing component width causes errors

dgutierr:
@ederign @kkufova Backport from master.


dgutierr:
@jhrcek 5 tests fails which are not related to this PR. wdyt?

jhrcek:
@dgutierr I wasn't planning to review this, I just executed full build for Klara to be ready. Now I see she's on PTO so I'll take a look and let you know within an hour hopefully..

dgutierr:
@jhrcek When 2 pages are opened at the same time, the resize event is captured by both instances causing columns with the same id to react. Notice that two pages with the same layout contain the same logical ids for its elements. 

Switching from id to hashCode forces the event to be captured only by the currently edited page (as expected). as hashCode GWT default implementation provides unique numbers for different objects.

dgutierr:
@ederign can we try to merge it before EOD please?

jhrcek:
@dgutierr makes sense. Thanks for explanation.

kkufova:

PR #298, opened 2018-04-16
AF-1139: Project migration tool does not preserve history of migrated…

mbarkley:
… repositories


mbarkley:
Related: https://github.com/kiegroup/kie-wb-common/pull/1643

mbarkley:
Jenkins, please retest this.

tomasdavidorg:
Hi @mbarkley,
I tested this version of migration tool and here are some results:
1. When .niogit directory is migrated and copied into the new workbench, the workbench don't index and load spaces and projects.
2. If you want to clone a migrated project via workbench ui, you will lost its history because of https://issues.jboss.org/browse/AF-702
3. If you clone a migrated repository (bare repo) via rest call, 'null' project is created, assets are not visible.
4. The only way how to get a new project to the workbench is to clone bare repo by command line and then clone it into workbench by rest call. In this case, I am able to see the history of assets.

mbarkley:
@tomasdavidorg Thanks for taking the time to test this.

The thing that concerns me is that most of the items you've listed seem unrelated to the changes in this PR. Could you please confirm which of these are problems are introduced by this PR?

tomasdavidorg:
@mbarkley sure... yes I think the problems are already in 7.7.x branch. I will check that.

mbarkley:
@tomasdavidorg The best thing to test in this PR would be running the migration tool on a repository containing multiple projects and verifying that the commit history is accurately preserved. Particularly, repositories with complex histories (i.e. merge commits, multiple authors, etc.) would be useful.

You should be able to run the migration tool on a multi-project repo without running any migration steps, and observe bare repos with the correct history as outputs.

tomasdavidorg:
Hi @mbarkley, 

I tested the migration tool again with several OU, repositories, projects and assets with some history and I didn't see any problem. The history migration seems to be correct. 

I also tried to reproduce the 1. problem I found but I am not able to reproduce it anymore. That is strange but I will test it once again when this PR is merge. It looks like I used some broken environment.

Thanks :-).

PR #312, opened 2018-04-23
AF-1155: Error importing projects located in a folder inside the repo

mbarkley:


mbarkley:
Related PR:
https://github.com/kiegroup/kie-wb-common/pull/1673

uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/PathUtil.java
```diff
@@ -51,4 +69,14 @@ public String getNiogitRepoPath(org.uberfire.java.nio.file.Path path) {
         }
     }
 
+    public Optional<String> extractBranch(String uri) {
+        final Pattern pattern = Pattern.compile("^[A-Za-z]+://([^@]+)@.*");
```
jhrcek: Pattern compilation is relatively expensive operation. Could we make this pattern into private static final field of the class?

uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/PathUtil.java
```diff
@@ -37,6 +40,21 @@
         return Paths.convert(path);
     }
 
+    public org.uberfire.backend.vfs.Path convert(org.uberfire.java.nio.file.Path path) {
+        return Paths.convert(path);
+    }
+
+    public String stripProtocolAndBranch(String uri) {
+        return uri.replaceFirst("^[A-Za-z]+://([^@]+@)?", "");
```
jhrcek: It would be nice to have these new methods unit tested.
Also `String#replaceFirst` is calling `Pattern.compile` internally. Based on how often we're expected to call this method, wouldn't it make sense to pre-compile the pattern in private static final field of this class?

tomasdavidorg:
Hi @mbarkley,

I tested this locally and I can see two issues:

1. If I clone a project with the new structure, I get dialog with error and Loading dialog. No error in server log.
screenshot from 2018-04-24 14-16-37

2. If I clone a repository with the old structure (several projects), I get Importing dialog.
screenshot from 2018-04-24 14-51-51

In both cases, the projects are cloned and indexed.

uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/util/commands/SyncRemote.java
```diff
@@ -76,6 +76,9 @@ public Optional execute() throws InvalidRemoteException {
             remoteBranches.removeAll(localBranches);
 
             for (final String branch : remoteBranches) {
+                if (branch.equals("HEAD")) {
```
jhrcek: You might want to use `org.eclipse.jgit.lib.Constants.HEAD` here. Just be aware we had a discussion about skipping branch creation for HEAD in [this PR](https://github.com/kiegroup/appformer/pull/243). For some reason @ederign advised that we should not skip it. Anyway it might be good idea to add comment about why we're skipping HEAD here (what problems it causes when not skipped).

tomasdavidorg:
Hi @mbarkley, the first issue is gone but I still can see the second one. The importing dialog remains open and the browser has to be refreshed to open library again.

tomasdavidorg:
@mbarkley I sent you email with steps to reproduce.

uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/util/commands/SyncRemote.java
mbarkley: Okay, I'll definitely change the code to use the jgit constant. @ederign can you explain why we wouldn't want to filter out the HEAD here? The reason I added this here is that without this line, cloned repositories would show the "HEAD" branch in the workbench UI.

mbarkley:
@tomasdavidorg Okay, I was able to reproduce issue (2) from the downstream war build. Going to rebuild locally with source maps and investigate.

mbarkley:
@tomasdavidorg I'm sorry for wasting your time, but I had made one more fix in kie-wb-common that I forgot to push. I've restarted the downstream build. You should be able to verify that (2) is fixed.

Note that when you import all the projects in bpm-assets some of the operations take a long time. It takes a while to import all the projects (though it does happen eventually) and when you are redirected to the Space screen, it takes a while before all the projects are listed (but also this does happen eventually). We might want to open a separate JIRA to fix this delay.

uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/util/commands/SyncRemote.java
ederign: I didn't read yet the whole thread, but my concern with removing the HEAD branch is that we could potentially lose the info if the HEAD points to another branch then master.

If the UI shows the HEAD branch I think is another issue.

ederign: I partially agree with you @mbarkley. 

My only concern for remove the HEAD branch will be when we will sync with an  repo/project. (something that we will probably add in future)

If a repo HEAD points to another branch/commit after they do the SyncRemote command, if we compare the local with remote will have potential differences.

I've two use cases for this:

- You want to fork my repo in the current branch in order to help me with some issue; (if we don't keep track of the head, the user will have to switch from master to this branch)
- Another option for this use case would be moving wb state to a detached head and ask maven to do a build from the detached head in order to try something. (How can you inform maven that he should build from a detached head?)

For sure this could be addressed in future but I only want to highlight this with you.

I'm good with both options, it's your call.

ederign: just to make it clear, I'm happy to merge this "as is".

mbarkley:
@jhrcek  @ederign I've decided to filter out HEAD, but add a comment explaining why. I've also used the jgit constant for HEAD.

jhrcek:
@mbarkley I think skipping HEAD is a good idea for now.

> Going to rebuild locally with source maps and investigate.

just FYI: sourcemaps are [enabled by default](https://github.com/kiegroup/kie-jenkins-scripts/blob/3da4c4345470660b7e7effbee21a923b8cab125d/job-dsls/jobs/downstream_pr_jobs.groovy#L13) for full downstream builds so you don't have to build locally - you can use FDB artifacts directly.

PR #322, opened 2018-04-27
DROOLS-2503 Switch Maven repository to repository.jboss.org.

pszubiak:
Signed-off-by: Piotr Szubiakowski <pszubiak@redhat.com>


porcelli:
@pszubiak please could you give us more context about this change?

pszubiak:
@porcelli it's related with jboss.org nexus performance. The problem is that we are skipping load balancers and sending requests direct to nexus web application. According to nexus team this is a reason of recent outages.

porcelli:
@pszubiak thx for the clarification.

manstis:
@pszubiak Do we need to update our local ```settings.xml``` to use the new URLs?

pszubiak:
@manstis Good point! Yes if you have origin-repository.jboss.org repository defined in local settings.xml file it would be good to change it to repository.jboss.org.

PR #331, opened 2018-05-02
JBPM-7167: [Stunner] Does not support save and rename feature

karreiro:
See https://issues.jboss.org/browse/JBPM-7167

Demo:
demo

---

Part of an ensemble:
- https://github.com/kiegroup/appformer/pull/331
- https://github.com/kiegroup/kie-wb-common/pull/1727

### DIFF
--- a/uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/BaseEditor.java
+++ b/uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/BaseEditor.java
@@ -648,16 +648,15 @@ public abstract class BaseEditor<T, M> {
     }
 
     boolean isContentDirty() {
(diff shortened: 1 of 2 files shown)


karreiro:
@manstis @jomarko Since we're defining content with a `null` value [here](https://github.com/karreiro/appformer/blob/af646c004c95d166988477ac0301451bc6c00c22/uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/test/java/org/uberfire/ext/editor/commons/client/BaseEditorTest.java#L515), a `NullPointerException` is raised [here](https://github.com/karreiro/appformer/blob/af646c004c95d166988477ac0301451bc6c00c22/uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/BaseEditor.java#L656).

I think that it's ok to keep the name of the test as is, because we're testing the `getCurrentContentHash` method `When` `ContentSupplierRaisesAnException` :-) wdyt?

karreiro:
@manstis Comment added :-)

PR #343, opened 2018-05-08
AF 1140: Add extension mechanism to enhance POM content

desmax74:
This PR adds an extension mechanism to enhance POM file content. Usefull to make sure you have the correct dependencies, repositories and plugins.
This is used by kie to enforce a valid pom file

### DIFF
--- a/uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/POMServiceImpl.java
+++ b/uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/POMServiceImpl.java
@@ -87,12 +87,14 @@ public class POMServiceImpl
                        final POM pomModel) {
         org.uberfire.java.nio.file.Path pathToPOMXML = null;
         try {
+            pomModel.addRepository(getRepository(repositoryWebBaseURL));
             final org.uberfire.java.nio.file.Path nioRoot = Paths.convert(projectRoot);
             pathToPOMXML = nioRoot.resolve("pom.xml");
+
             if (ioService.exists(pathToPOMXML)) {
                 throw new FileAlreadyExistsException(pathToPOMXML.toString());
             }
-            Model model = getUpdatedPom(pathToPOMXML, repositoryWebBaseURL);
+            final Model model = pomEnhancer.execute(pomContentHandler.convert(pomModel));
             write(model, pathToPOMXML, ioService);
             //Don't raise a NewResourceAdded event as this is handled at the Project level in ProjectServices
             return Paths.convert(pathToPOMXML);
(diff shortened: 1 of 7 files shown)


uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/POMServiceImpl.java
```diff
@@ -45,6 +55,9 @@
     private POMContentHandler pomContentHandler;
     private M2RepoService m2RepoService;
     private MetadataService metadataService;
+    private MavenXpp3Reader reader;
```
jhrcek: I don't see this `reader` field used anywhere - do we need it?

uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/POMServiceImpl.java
```diff
@@ -97,6 +111,25 @@ private MavenRepository getRepository(final String baseURL) {
...
+    private boolean write(Model model, org.uberfire.java.nio.file.Path pathToPOMXML, IOService ioService) {
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        try {
+            writer.write(baos, model);
+            ioService.write(pathToPOMXML, new String(baos.toByteArray(), StandardCharsets.UTF_8));
+            return true;
+        } catch (Exception e) {
+            return false;
+        } finally {
+            try {
```
jhrcek: This bit of stream closing boilerplate can be removed if you use [try-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html), which closes the stream for you:

```
        try (ByteArrayOutputStream baos = new ByteArrayOutputStream()) {
            writer.write(baos, model);
            ioService.write(pathToPOMXML, new String(baos.toByteArray(), StandardCharsets.UTF_8));
        } catch (Exception e) {

        }
```

uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/POMServiceImpl.java
```diff
@@ -97,6 +111,25 @@ private MavenRepository getRepository(final String baseURL) {
...
     }
 
+    private boolean write(Model model, org.uberfire.java.nio.file.Path pathToPOMXML, IOService ioService) {
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        try {
+            writer.write(baos, model);
+            ioService.write(pathToPOMXML, new String(baos.toByteArray(), StandardCharsets.UTF_8));
+            return true;
+        } catch (Exception e) {
+            return false;
```
jhrcek: It might be good to at least log the exception so users know that pom migration failed.

uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/POMServiceImpl.java
```diff
@@ -97,6 +111,25 @@ private MavenRepository getRepository(final String baseURL) {
         return mavenRepository;
     }
 
+    private boolean write(Model model, org.uberfire.java.nio.file.Path pathToPOMXML, IOService ioService) {
```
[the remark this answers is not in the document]
desmax74: Before was used now is not readed , changed to void

uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/POMServiceImpl.java
desmax74: Was here due to the move of the code, but here is not needed, removed

uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/POMServiceImpl.java
desmax74: Thanks for the tips :)

uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/POMServiceImpl.java
desmax74: Logged

PR #359, opened 2018-05-16
JBPM-5813: Stunner - Open XML editor instead of sorry screen when process can't be opened

manstis:
Ensure BaseFileUploadServlet creates the file in a batch.

See https://issues.jboss.org/browse/JBPM-5813 (more specifically the reported issue on the Product [JIRA](https://issues.jboss.org/browse/RHPAM-447?focusedCommentId=13576647&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13576647)).

### DIFF
--- a/uberfire-server/src/main/java/org/uberfire/server/BaseUploadServlet.java
+++ b/uberfire-server/src/main/java/org/uberfire/server/BaseUploadServlet.java
@@ -18,6 +18,7 @@ package org.uberfire.server;
 
 import java.io.IOException;
 import java.util.Iterator;
+
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
@@ -64,14 +65,19 @@ public abstract class BaseUploadServlet extends BaseFilteredServlet {
     protected void writeFile(final IOService ioService,
                              final Path path,
                              final FileItem uploadedItem) throws IOException {
-        if (!ioService.exists(path)) {
-            ioService.createFile(path);
+        try {
+            ioService.startBatch(path.getFileSystem());
+
+            if (!ioService.exists(path)) {
+                ioService.createFile(path);
+            }
+
+            ioService.write(path,
+                            IOUtils.toByteArray(uploadedItem.getInputStream()));
+        } finally {
+            uploadedItem.getInputStream().close();
+            ioService.endBatch();
         }
-
-        ioService.write(path,
-                        IOUtils.toByteArray(uploadedItem.getInputStream()));
-
-        uploadedItem.getInputStream().close();
     }
 
     protected void logError(Throwable e) {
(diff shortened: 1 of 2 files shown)


manstis:
Explanation of the issue can be found [here](https://issues.jboss.org/browse/RHPAM-447?focusedCommentId=13577472&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13577472).

manstis:
@paulovmr Could you please review from an Engineering perspective (in the absence of others).

uberfire-server/src/main/java/org/uberfire/server/BaseUploadServlet.java
```diff
@@ -64,14 +65,21 @@ protected ServletFileUpload getServletFileUpload() {
...
-                        IOUtils.toByteArray(uploadedItem.getInputStream()));
+            if (!ioService.exists(path)) {
+                ioService.createFile(path);
+            }
 
-        uploadedItem.getInputStream().close();
+            ioService.write(path,
+                            IOUtils.toByteArray(uploadedItem.getInputStream()));
+        } catch (final Exception e) {
+            throw new RuntimeException(e);
```
jomarko: I don't see benefit wrapping any `Exception` to `RuntimeExcetpion`, could you explain please?

manstis: @jomarko It was copy and paste demons! I've now simply wrapped the ```ioService``` operations in a ```try {..} finally {..}``` block... that is also in keeping with how error handling was before my change; so less likely to introduce any regressions in any existing error handling.

jomarko:
@manstis Correct me if I am wrong, For checking I should use Import Asset dialog, right? The problem is it remains opened even after successful import, see the attached import, could you please check if it is or it is not caused by your changes?
screenshot from 2018-05-17 12-26-04

manstis:
@jomarko When I run Stunner Showcase (Project) with the changes after uploading a file the "Upload dialog" disappears and I am taken to the Stunner BPMN Designer. What webapp did you use? (Happy to try the same here).

manstis:
Hi @jomarko I just completed trying ```kie-drools-wb``` 7.8.0-SNAPSHOT and it worked OK.

You need to rename the "broken" files to ```.bpmn``` to open Stunner. Keeping ```.bpmn2``` opens ```jbpm-designer```... however I tried with both extensions and the respective editors opened and the "Upload" dialog disappeared.

jomarko:
@manstis what should be behavior if user opens a broken bpmn file in xml editor, fix the issues in file and save the file? Currently xml editor remains opened. Not saying it is wrong, just want to be sure if the process designer shouldn't be shown automatically once broken file fixed and saved. Currently user has to reopen the bpmn file after the fix is saved to get process designer opened.

manstis:
@jomarko As you've noticed, at the moment the editor does not re-open as "Stunner" when the XML is corrected. The check for "parsing errors" happens when the file is loaded. An error results in the client being instructed to open the XML Editor; success the Stunner Editor. The expected behaviour is undefined so what we have now is probably acceptable.

PR #361, opened 2018-05-17
RHPAM-1017 Add freemarker into dependencyMgmt

ryanzhang:
This is for the same reason as https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/757


mareknovotny:
ok to test

PR #364, opened 2018-05-22
AF-1205 : Use system property for default m2 repository

Rikkola:
https://issues.jboss.org/browse/AF-1205

### DIFF
--- a/uberfire-m2repo-editor/uberfire-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/M2RepoServiceImpl.java
+++ b/uberfire-m2repo-editor/uberfire-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/M2RepoServiceImpl.java
@@ -27,6 +27,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
+
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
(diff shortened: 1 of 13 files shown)


uberfire-m2repo-editor/uberfire-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/M2RepoServiceImpl.java
```diff
@@ -232,20 +233,17 @@ GAV getGAV(final String path) {
...
+
+        if (urlProperty == null || urlProperty.isEmpty()) {
             return repository.getRepositoryURL(ArtifactRepositoryService.GLOBAL_M2_REPO_NAME);
         } else {
-            if (baseURL.endsWith("/")) {
-                return baseURL + "maven2/";
-            } else {
-                return baseURL + "/maven2/";
-            }
+            return urlProperty;
```
jhrcek: Would it make sense to add some error checking, it the string from the system property constitutest valid url?
Like createing `java.net.URL(String)` and catching `MalformedUrlException` and then either rethrowing or returning some default..

Rikkola: Sure we can do that.

Rikkola:
@jhrcek sure making the docs at the moment. Updating the properties section and N&N.

@ederign Hi, would you have time to review this and https://github.com/kiegroup/appformer/pull/366?

PR #368, opened 2018-05-23
JBPM-7245  Diagram SVG is not highlighted on the WB

tiagodolphine:
Changes to support Diagram SVG is not highlighted on the WB (jbpm-process-svg widget).
Insert the node id in the same way it was done for the containers.

PS: unit tests will be updated later, the PR is to be tested ASAP since this is a blocker.

@romartin 
@LuboTerifaj


romartin:
@ederign I merged this without asking sorry, but it's reaaaaly urgent.... we love you <3

PR #370, opened 2018-05-23
[7.7.x] AF-1208 : Migrated project by pom migration tool can't be cloned into BC via REST clone

Rikkola:
(cherry picked from commit a800badcf5f3921e8b4fc94825a4f5828a08df66)

https://issues.jboss.org/browse/AF-1208


uberfire-rest/uberfire-rest-backend/src/main/java/org/guvnor/rest/backend/JobRequestHelper.java
```diff
@@ -115,6 +115,9 @@ public JobResult cloneProject(final String jobId,
         if (cloneProjectRequest.getPassword() != null && !"".equals(cloneProjectRequest.getPassword())) {
             configuration.setPassword(cloneProjectRequest.getPassword());
         }
+
+        configuration.setInit(false);
+        configuration.setMirror(false);
```
Rikkola: I copied this setup from the UI repository import. So we already use them in other location.

Rikkola: import here = clone

PR #395, opened 2018-06-07
[AF-1238] FileSystem Watch Events should reproduce all the history on BatchMode

ederign:
### **Problem**

The story begins [here](https://issues.jboss.org/browse/RHPAM-559
). 
Data modeler code does the following: 

1. create a file with content (A.java);
2. Rename this file;
	2.1 Start a batch();
    2.2 Rename this file A.java -> B.java (one git commit)
    2.3 Change the content of this file in the refactoring process; (one git commit)
    2.4 Close batch 

The issue is that instead of a renaming(a->b) + modify event(b) they are receiving a delete on a and created in b.

@wmedvede did a great job helping me isolating this issue and what happening is that our VFS in batch mode does a amend of each commit and in the end batch process compare the three in 2.1 with the tree in 2.2 and get the 'wrong' events (or at least a compressed vision of the timeline).

The problem is that on git, when I do the same in the CLI:

- git diff HEAD HEAD~1 After 2.2
```
diff --git a/B.java b/A.java
similarity index 100%
rename from B.java
rename to A.java
(END)
```

- git diff HEAD HEAD~1 After 2.3
```
diff --git a/B.java b/B.java
index 47caeb6..cbfee01 100644
--- a/B.java
+++ b/B.java
@@ -1 +1 @@
- fsdljdssklfsjdfsdfjsdl
+ fsdfjsdl
```
When I do a git rebase -i HEAD~2 (like our batch does)
- git diff HEAD HEAD~1
```
diff --git a/A.java b/A.java

new file mode 100644
index 0000000..cbfee01
--- /dev/null
+++ b/A.java
@@ -0,0 +1 @@
+ fsdfjsdl

diff --git a/B.java b/B.java
deleted file mode 100644
index 47caeb6..0000000
--- a/B.java
+++ /dev/null
@@ -1 +0,0 @@
- fsdljdssklfsjdfsdfjsdl
```

See the history? Where is the rename/move? 🔪 😭 

Git does some magic or lost himself when comparing small/new files amended.

### **Proposed Solution**

Instead of comparing the start/end tree in the end of batch process I compared the tree in each commit and stored all the events associated. In our case we will have a rename event + a modify event. Easy isn't? 🤔 

### **Drawbacks**
That is where I'll need help from @manstis, @porcelli and @adrielparedes.

With the current/compressed behaviour, If I do something like:

1. Start a batch();
2. create b
3. modify b
4. modify b
5. modify b
6. modify b
2. End  batch

I would get only one B created event (remember that the commits are amended). Triggered the index/build just once.

With this new reality I would get 5 events. 😭 

This could be even worse:

1. Start a batch();
2. create b
3. modify b
4. modify b
5. modify b
6. modify b
7. delete b
2. End  batch

No event triggered. With the new solution we will have 6 events! 👍 

### ***Conclusion(?)***
I'm almost sure that our VFS should create all the events for all the changes, even if those changes are amended, but we will  need for sure to consolidate the events on [IOServiceIndexedImpl.](https://github.com/kiegroup/appformer/blob/69a58e6f5d699f0b6a4e957c89f7b6f28311ccb4/uberfire-extensions/uberfire-metadata/uberfire-metadata-commons-io/src/main/java/org/uberfire/ext/metadata/io/IOServiceIndexedImpl.java#L324) in short future (or event before merging this PR).

What do you guys think?

### DIFF
--- a/uberfire-io/src/test/java/org/uberfire/io/impl/BatchTest.java
+++ b/uberfire-io/src/test/java/org/uberfire/io/impl/BatchTest.java
@@ -37,12 +37,14 @@ import org.uberfire.java.nio.base.options.CommentedOption;
 import org.uberfire.java.nio.base.version.VersionAttributeView;
 import org.uberfire.java.nio.file.FileSystem;
 import org.uberfire.java.nio.file.Path;
+import org.uberfire.java.nio.file.StandardWatchEventKind;
 import org.uberfire.java.nio.file.WatchEvent;
 import org.uberfire.java.nio.file.WatchService;
 import org.uberfire.java.nio.file.api.FileSystemProviders;
 import org.uberfire.java.nio.fs.jgit.JGitFileSystemImpl;
 import org.uberfire.java.nio.fs.jgit.JGitFileSystemProvider;
 import org.uberfire.java.nio.fs.jgit.JGitFileSystemProxy;
+import org.uberfire.java.nio.fs.jgit.ws.JGitWatchEvent;
 
 import static org.junit.Assert.*;
 import static org.mockito.Mockito.*;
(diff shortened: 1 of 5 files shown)


manstis:
@porcelli @ederign I think we effectively consolidate paths [here](https://github.com/kiegroup/appformer/blob/master/uberfire-extensions/uberfire-metadata/uberfire-metadata-commons-io/src/main/java/org/uberfire/ext/metadata/io/IOServiceIndexedImpl.java#L416). Although I'm less certain if we should also consolidate "delete" events too?

ederign:
Exactly @manstis.

Porcelli, we know that we have those situations on git and we should address them right?

Also, I don't think we have right now these situations that I've mentioned (i.e. created file x, change x, delete x) on the same batch on the workbench anyway right?

So, I'm not sure that we will have this impact on the workbench performance, but for sure we should be smarter on indexing.

What do you guys think of merging this because it fixes the git behavior the Data modeller issue and fixes (if necessary) the indexing on another JIRA?

Or should we wait for someday we fix the index and merge this?

wmedvede:
@porcelli  @ederign  @manstis  I'd vote for keep this it as is for the 7.0.2 and with more time do the change for the 7.1.0.
Why?
1) The issue reported right now for the data modeler happens only when the user renames a data object in the "Project Explorer". When the user renames the data object by using then "Rename" button, it's working fine.
2) the issue was reported by QE and not by a customer, so until now there are no customers waiting for this fix. It's also a "Major" issue (not blocker nor critical), and with the library stuff I believe users are really not using the project explorer for renaming files by using the repository view.
3) all the other editors has worked until now, and AFAIK there are no other reports related with this.

Regards,
Walter.

porcelli:
+1 for @wmedvede rationale

ederign:
@porcelli so we are good to have this on master?

porcelli:
if @manstis and @adrielparedes +1.. i'm good with that :)

adrielparedes:
+1

wmedvede:
Hi @ederign I'm not expert on this area but looks good to me. I've built and tested this stuff locally and now we have that two events are produced as part of the batch processing scenario we were analyzing together.

1) a ResourceRenamedEvent (#time 1), for path1
2) a ResourceUpdatedEvent   (#time 2), for path2

See that the chronological order is #time1, #time2 , this is fine, however some thing that comes to my mind right now is that when events produced by the WatchService are iterated, see line 99, we basically process this events in chronological order. This is good. 
See that this two events are for different paths. 

But, then at some point of the processing in line 110 this events are being stored in a final Map<Path, Collection<ResourceChange>> changes. So we basically store the events in a Map by indexing by path.

And this Map of changes will be finally fired in line 122 in a ResourceBatchChangesEvent. (line 122 is not in the picture)

Finally interested parties usually Observes the ResourceBatchChangesEvent, typically "iterates" the Map entries, etc.
But since the events are not stored in the Map by "Path", it could be the case where the chronological order #time1, #time2 is lost.
Probably a different structure than a Map should be returned, e.g. just a List of events populated in the order we process the events produced by the WatchService. An just let the consumers to iterate in the chronological order and pick what they want.
WDYT?

p.s. I know this Map of events has been this way since long time ago and things has worked, but now that I'm looking into this question come to mi mind.

eventsorder

manstis:
I'm not around much at the moment; happy to go with @wmedvede rationale.

Please do not wait for my comments/opinion as it's likely to delay anything for weeks.

PR #407, opened 2018-06-29
JBPM-7404: Folder related to a new space creation is not created until

jethromx:
first project is created

### DIFF
--- a/uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/organizationalunit/OrganizationalUnitServiceImpl.java
+++ b/uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/organizationalunit/OrganizationalUnitServiceImpl.java
@@ -15,6 +15,7 @@
 
 package org.guvnor.structure.backend.organizationalunit;
 
+import java.io.File;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/organizationalunit/OrganizationalUnitServiceImpl.java
```diff
@@ -225,7 +228,19 @@ public OrganizationalUnit createOrganizationalUnit(final String name,
             newOrganizationalUnit = organizationalUnitFactory.newOrganizationalUnit(groupConfig);
             registeredOrganizationalUnits.put(newOrganizationalUnit.getName(),
                                               newOrganizationalUnit);
-
+           
+            ConfigProperties configProperties= new ConfigProperties(System.getProperties());
+            JGitFileSystemProviderConfiguration jGitFileSystemProviderConfiguration =new JGitFileSystemProviderConfiguration();
```
ederign: This is a lack of abstraction. OrganizationUnit/space is only an abstract concept and shouldn't have any knowledge of jgit internals and also it's not their responsibility to create directories on .niogit.

PR #410, opened 2018-07-05
[7.7.x] RHPAM-832: Dashbuilder tests failing on windows

dgutierr:
@hasys @kkufova backport from master.

### DIFF
--- a/dashbuilder/dashbuilder-backend/dashbuilder-dataset-cdi/src/test/java/BootstrapTest.java
+++ b/dashbuilder/dashbuilder-backend/dashbuilder-dataset-cdi/src/test/java/BootstrapTest.java
@@ -28,6 +28,7 @@ import org.dashbuilder.dataset.json.DataSetDefJSONMarshaller;
 import org.dashbuilder.test.BaseCDITest;
 import org.jboss.arquillian.junit.Arquillian;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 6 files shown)


dgutierr:
@ederign Can you ack please and merge once it's approved by QE? @jhrcek  ^

dashbuilder/dashbuilder-backend/dashbuilder-dataset-cdi/src/test/java/BootstrapTest.java
```diff
@@ -37,6 +38,7 @@
 import static org.mockito.Mockito.*;
 
 @RunWith(Arquillian.class)
+@Ignore // see https://issues.jboss.org/browse/RHPAM-832
```
jhrcek: Does it make sense to keep these tests if they're never going to be executed anyway? If there's no work planned for unignoring these we might as well delete these, because nobody will ever fix these.

If you insist on keeping the tests and just having those ignored I'd suggest:
1. add reason for ignoring as `value` of the annotation (to make it accessible to tooling) instead of just plain comment. I.e. use `@Ignore{"bla"}` instead of `@Ignore //bla`
2. Add either inline comment or info to the jira stating when these could be unignored or what further steps / investigations need to be done to make it possible to unignore these.

dgutierr: Done 1 & 2.

domhanak:
@dgutierr Can we merge this?

PR #417, opened 2018-07-12
AF-799

desmax74:
It contains configuration classes for the compiler and a constructor change in one class to be usable, the code is only on the maven compiler and not the integration of the workbench,
it depends on:
https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/781

@jhrcek @porcelli @jakubschwan

### DIFF
--- a/uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/configuration/ConfigurationKey.java
+++ b/uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/configuration/ConfigurationKey.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.guvnor.common.services.project.backend.server.utils.configuration;
+
+public enum ConfigurationKey {
+
+    COMPILER,
+    SOURCE_VERSION,
+    TARGET_VERSION,
+    FAIL_ON_ERROR,
+
+    MAVEN_COMPILER_PLUGIN_GROUP,
+    MAVEN_COMPILER_PLUGIN_ARTIFACT,
+    MAVEN_COMPILER_PLUGIN_VERSION,
+
+    TAKARI_COMPILER_PLUGIN_GROUP,
+    TAKARI_COMPILER_PLUGIN_ARTIFACT,
+    TAKARI_COMPILER_PLUGIN_VERSION,
+
+    KIE_MAVEN_PLUGINS,
+    KIE_MAVEN_PLUGIN,
+    KIE_TAKARI_PLUGIN,
+
+    KIE_VERSION;
+}
--- a/uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/configuration/ConfigurationKey.java
+++ b/uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/configuration/ConfigurationKey.java
@@ -22,7 +22,7 @@ package org.guvnor.common.services.project.backend.server.utils.configuration;
  * <p>
  *  COMPILER (jdt) jdt or javac
  *  SOURCE_VERSION  (1.8) configured with jvm source code version
- *  TARGET_VERSION  (1.8)configured with jvm target version
+ *  TARGET_VERSION  (1.8) configured with jvm target version
  *  <p>
  *  FAIL_ON_ERROR (false) configured with false to continue the build on the correct classes and skip the build of classes with errors
  *  MAVEN_COMPILER_PLUGIN_GROUP (org.apache.maven.plugins) configured with default maven compiler group to disabled it
--- a/uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/configuration/ConfigurationKey.java
+++ b/uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/configuration/ConfigurationKey.java
@@ -17,27 +17,27 @@
 package org.guvnor.common.services.project.backend.server.utils.configuration;
 
 /**
- * Keys use to configure the Maven compiler in the kie-wb-common
+ * Keys used to configure the Maven compiler in the kie-wb-common
  *
  * <p>
- *  COMPILER jdt or javac
- *  SOURCE_VERSION  configured with jvm source code version
- *  TARGET_VERSION  configured with jvm target version
+ *  COMPILER (jdt) jdt or javac
+ *  SOURCE_VERSION  (1.8) configured with jvm source code version
+ *  TARGET_VERSION  (1.8)configured with jvm target version
  *  <p>
- *  FAIL_ON_ERROR configured with false to continue the build on the correct classes and skip the build of classes with errors
- *  MAVEN_COMPILER_PLUGIN_GROUP  configured with default maven compiler group to disabled it
- *  MAVEN_COMPILER_PLUGIN_ARTIFACT configured with default maven compiler ArtifactID to disabled it
- *  MAVEN_COMPILER_PLUGIN_VERSION configured with default maven compiler version to disabled it
+ *  FAIL_ON_ERROR (false) configured with false to continue the build on the correct classes and skip the build of classes with errors
+ *  MAVEN_COMPILER_PLUGIN_GROUP (org.apache.maven.plugins) configured with default maven compiler group to disabled it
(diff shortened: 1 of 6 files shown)


uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/configuration/ConfigurationKey.java
```diff
@@ -0,0 +1,40 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.guvnor.common.services.project.backend.server.utils.configuration;
+
+public enum ConfigurationKey {
```
porcelli: javadoc?

desmax74: Fixed

uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/configuration/ConfigurationKey.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Keys use to configure the Maven compiler in the kie-wb-common
+ *
+ * <p>
+ *  COMPILER jdt or javac
+ *  SOURCE_VERSION  configured with jvm source code version
+ *  TARGET_VERSION  configured with jvm target version
+ *  <p>
+ *  FAIL_ON_ERROR configured with false to continue the build on the correct classes and skip the build of classes with errors
+ *  MAVEN_COMPILER_PLUGIN_GROUP  configured with default maven compiler group to disabled it
```
jhrcek: I don't understand this comment `configured with default maven compiler group to disabled it` - what do you mean by it? What do I need to disable and why? When would it make sense to enable it?

uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/configuration/ConfigurationKey.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+package org.guvnor.common.services.project.backend.server.utils.configuration;
+
+/**
+ * Keys use to configure the Maven compiler in the kie-wb-common
+ *
+ * <p>
+ *  COMPILER jdt or javac
+ *  SOURCE_VERSION  configured with jvm source code version
```
jhrcek: What's the expected version format (1.8 versus 8)? Which versions do we want to support?
Please add an example of the expected version format to the javadoc.

uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/configuration/ConfigurationKey.java
```diff
@@ -0,0 +1,62 @@
...
+ *  FAIL_ON_ERROR configured with false to continue the build on the correct classes and skip the build of classes with errors
+ *  MAVEN_COMPILER_PLUGIN_GROUP  configured with default maven compiler group to disabled it
+ *  MAVEN_COMPILER_PLUGIN_ARTIFACT configured with default maven compiler ArtifactID to disabled it
+ *  MAVEN_COMPILER_PLUGIN_VERSION configured with default maven compiler version to disabled it
+ *  <p>
+ *  TAKARI_COMPILER_PLUGIN_GROUP configured with takari GroupID
+ *  TAKARI_COMPILER_PLUGIN_ARTIFACT configured with takari ArtifactID
+ *  TAKARI_COMPILER_PLUGIN_VERSION configured with a placeholder and set with the correct value in the maven build with takari version
+ *  <p>
+ *  KIE_MAVEN_PLUGINS configured with the GroupID of kie plugin from the Integration prj
```
jhrcek: The name of this property seems inconsistent with the javadoc comment - shouldn't it have GROUP/GROUPS in the name?

uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/configuration/Order.java
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.guvnor.common.services.project.backend.server.utils.configuration;
+
+/***
+ * Used to order execution of implementations of the same interface in
```
jhrcek: Looks like unfinished sentence..

uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/configuration/ConfigurationStrategy.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.Map;
+
+/**
+ * Define the behaviour of a ConfigurationStrategy,
+ * load the configuration and check if it's valid
+ */
+public interface ConfigurationStrategy extends Valid,
+                                               Order {
+
+    Map<ConfigurationKey, String> loadConfiguration();
```
jhrcek: This design of having separate `loadConfiguration` and `isValid` methods does not enforce correct usage of the API. For example user can load the configuration which is invalid and then forget to call `isValid()` and do something about invalid configurationd.

In my opinion ConfigurationStrategy should only have one method like 
`Map<ConfigurationKey, String> loadConfiguration() throws InvalidConfigurationException;` where `InvalidConfigurationException` would be checked exception, forcing the user to handle it.
Also that exception could hold details about what is invalid like: missing key? invalid/unparseable value of property? Invalid / unsupported combination of properties?

@desmax74 what do you think?

uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/configuration/ConfigurationKey.java
```diff
@@ -0,0 +1,62 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.guvnor.common.services.project.backend.server.utils.configuration;
+
+/**
+ * Keys use to configure the Maven compiler in the kie-wb-common
```
jhrcek: Typo: use -> used

uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/configuration/ConfigurationKey.java
desmax74: 1.8,  support older version requires tests and the availability of versions currently not freely available. I'll add the current values in the javadoc.

uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/configuration/ConfigurationKey.java
desmax74: Added Group and Artifact on the KIE_MAVEN_PLUGIN_ and on KIE_TAKARI_PLUGIN

uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/configuration/Order.java
desmax74: Fixed

uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/configuration/ConfigurationStrategy.java
desmax74: The isValid is used by the ConfigurationContextProvider to understand if is ok the Strategy impl or if is better to use the next, using the exceptions to understand if is valid or not in this internal API isn't a good idea because a user will be forced to use a configuration not wanted to remove error messages.

uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/configuration/ConfigurationKey.java
desmax74: Fixed

uberfire-project/uberfire-project-backend/src/main/java/org/guvnor/common/services/project/backend/server/utils/configuration/ConfigurationKey.java
```diff
@@ -0,0 +1,62 @@
...
+
+package org.guvnor.common.services.project.backend.server.utils.configuration;
+
+/**
+ * Keys used to configure the Maven compiler in the kie-wb-common
+ *
+ * <p>
+ *  COMPILER (jdt) jdt or javac
+ *  SOURCE_VERSION  (1.8) configured with jvm source code version
+ *  TARGET_VERSION  (1.8)configured with jvm target version
```
jakubschwan: just nitpick. missing space after end of the column

desmax74: Fixed

desmax74:
@ederign

PR #422, opened 2018-07-24
AF-1412 : Kie WB REST getJob randomly return 410 Gone after previous REST call

Rikkola:
https://issues.jboss.org/browse/AF-1412

Bundle:
https://github.com/kiegroup/appformer/pull/422
https://github.com/kiegroup/kie-wb-distributions/pull/782

### DIFF
--- a/uberfire-rest/uberfire-rest-backend/src/main/java/org/guvnor/rest/backend/ProjectResource.java
+++ b/uberfire-rest/uberfire-rest-backend/src/main/java/org/guvnor/rest/backend/ProjectResource.java
@@ -78,7 +78,12 @@ import static org.guvnor.rest.backend.PermissionConstants.REST_ROLE;
 public class ProjectResource {
 
     private static final Logger logger = LoggerFactory.getLogger(ProjectResource.class);
-    private static Variant defaultVariant = Variant.mediaTypes(MediaType.APPLICATION_JSON_TYPE).add().build().get(0);
+    private Variant defaultVariant = getDefaultVariant();
+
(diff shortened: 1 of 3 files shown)


Rikkola:
Had to change few methods to protected for testability. This is since the use of `Variant` gives class not found errors and fixing them looked messier.

uberfire-rest/uberfire-rest-backend/src/main/java/org/guvnor/rest/backend/ProjectResource.java
```diff
@@ -103,6 +106,7 @@ private void addAcceptedJobResult(String jobId) {
         JobResult jobResult = new JobResult();
         jobResult.setJobId(jobId);
         jobResult.setStatus(JobStatus.ACCEPTED);
+        jobManager.putJob(jobResult);
```
jhrcek: @Rikkola The 2 significant changes I see is this line and removal of `@Inject SpaceAPI`. Could you please explain how that fixes the issue? Also now (provided that you think this should fix the issue) we probably don't need [this workaround](https://github.com/kiegroup/kie-wb-distributions/pull/757) - could you please remove it to make the tests a bit faster? (CC @cristianonicolai just so he knows)

Rikkola: @jhrcek When I wrote the test for this I noticed the file had the SpaceAPI injected twice. It doesn't actually fix anything, it was just not used.

And this line was accidentally removed during the Project Orientated changes. So added it back.

PR #424, opened 2018-07-27
upgraded src/main/config/zanata.xml to link to new project 7.10.0 on https://vendors.zanata.redhat.com/

kiereleaseuser:


src/main/config/zanata.xml
```diff
@@ -3,7 +3,7 @@
 <config xmlns="http://zanata.org/namespace/config/">
   <url>https://vendors.zanata.redhat.com/</url>
   <project>uberfire</project>
-  <project-version>2.6.0</project-version>
+  <project-version>2.7.0</project-version>
```
manstis: Isn't commit message wrong? You're upgrading to 2.7.0

kiereleaseuser: this is the thing I don't like - uberfire and dashbuilder are mantaining their initial version - not like all other kie-projects 7.10.0

kiereleaseuser: I see the commit msg is weird - that's because of the script  - it does a gitall add . and gitall commit with the same commit message - have to change this

kiereleaseuser: @manstis ^^

kiereleaseuser: @manstis script changed - so the next time the commit msg for appformer will be right

PR #425, opened 2018-07-30
[AF-1432] Remove AF dependencies from Uberfire Core Grids component

danielezonca:
Removed all UF/WB related dependencies to Uberfire Core Grids component to become a pure Lienzo/Gwt component and maybe moved to an independent Lienzo-Grid component in the future.

- Removed `PortablePreconditions` usages
- Changed `Callback` to `Consumer`
- Changed `Command` to `Runnable`
- Changed `Pair` to `Map.Entry`

**NOTE:** I used `Objects` methods to replace `PortablePreconditions`. This changes the type of exception fired in case of null-check error: before was an `IllegalStateException` and now a `NullPointerException`. If needed I can implement an alternative to restore the same behaviour but I would like to use a built-in mechanism instead of a custom one.

@manstis @mdproctor @romartin @jsoltes 

PRs related:
https://github.com/kiegroup/appformer/pull/425
https://github.com/kiegroup/drools-wb/pull/902
https://github.com/kiegroup/kie-wb-common/pull/2009

### DIFF
--- a/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/model/impl/BaseGridData.java
+++ b/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/model/impl/BaseGridData.java
@@ -15,6 +15,7 @@
  */
 package org.uberfire.ext.wires.core.grids.client.model.impl;
 
+import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -26,8 +27,6 @@ import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.IntStream;
 
-import org.kie.soup.commons.validation.PortablePreconditions;
-import org.uberfire.commons.data.Pair;
 import org.uberfire.ext.wires.core.grids.client.model.GridCell;
 import org.uberfire.ext.wires.core.grids.client.model.GridCellValue;
 import org.uberfire.ext.wires.core.grids.client.model.GridColumn;
@@ -273,8 +272,9 @@ public class BaseGridData implements GridData {
 
     @Override
     public void setHeaderRowCount(final int headerRowCount) {
-        PortablePreconditions.checkCondition("headerRowCount",
-                                             headerRowCount > 0);
+        if (!(headerRowCount > 0)) {
+            throw new IllegalArgumentException("headerRowCount");
+        }
         this.headerRowCount = headerRowCount;
     }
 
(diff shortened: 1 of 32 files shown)


danielezonca:

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/model/impl/BaseGridData.java
```diff
@@ -397,7 +397,7 @@ protected Range doSetCell(final int rowIndex,
         //If we're not merged just set the value of a single cell
         if (!isMerged) {
             ((BaseGridRow) rows.get(rowIndex)).setCell(_columnIndex,
-                                                       cellSupplier.apply(Pair.newPair(rowIndex, columnIndex)));
+                                                       cellSupplier.apply(new AbstractMap.SimpleEntry<>(rowIndex, columnIndex)));
```
jsoltes: @danielezonca what was the reason for changing Pair with Map.Entry? Just curious...

danielezonca: @jsoltes 
The goal was to remove the dependency with `org.uberfire.commons.data.Pair` to isolate grid component and I just decided to use a built-in class instead of create a new bean just to wrap two values. No other specific/technical reason :)

jsoltes: @danielezonca ok, I struck me right after I wrote it. I thought it was the javafx.util.Pair, didn't realize the connection to uberfire at first.

danielezonca:
I'm looking at failed (19) tests but I think they are not related to this PR.
- 2 of them are related to DMN and OSGI and it is a known bug with fix in progress
- 1 is a Drools integration error and I don't see any useful info in the exception so maybe it was just a timeout
- 13 are jBPM backend integration tests so I don't really think can have an impact from an AppFormer PR (some of the errors are timeout or similar)
- 2 are in kie-wb-distribution and seem timeout too (a generic `null` as message) 
- 1 is related to WB maven compilation but I checked with @desmax74 and is not related to this PR 

So I will ask another build to see if were random failures

danielezonca:

ederign:
@danielezonca sorry for the delay on this review, I'll take a look on this tomorrow morning.

jsoltes:
@danielezonca For recognizing the random failures, you might find this webpage useful: http://janhrcek.cz/random-failures/ And if you look at it maybe another rebuild won't be necessary.

danielezonca:
Ok now with the same codebase failures are 68...

@jsoltes Thanks for the link, I checked the failures of previous run and all of them are present in that list. I don't know now the status of those 68 but I think are random failure too

Let's wait @ederign review but I think we can merge

PR #428, opened 2018-08-03
JBPM-7537: [GSS] (7.0.z) Cannot use multibyte class name for Data Objects

wmedvede:


uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/io/watch/IOWatchServiceExecutorImpl.java
```diff
@@ -111,7 +109,7 @@ public void execute(final WatchKey wk,
                     }
                 }
             }
-            if (changes.size() == 1 && changes.values().size() == 1) {
+            if (changes.size() == 1 && changes.entrySet().iterator().next().getValue().size() == 1) {
```
wmedvede: Hi @ederign , this line here introduces a quick win managing events for the same file on the same batch processing.
Before: 
if FileA was e.g.
    1) Updated
    2) Renamed
only the first produced event was processed, and a single ResourceUpdatedEvent was raised. But this was wrong, since two events were produced.

Now with the new condition
changes.size() == 1 && changes.entrySet().iterator().next().getValue().size() == 1
we can detect that two events were produced, and thus a ResourceBatchChangesEvent will be produced as expected.

wmedvede:
I've added the comment above but forgot ask you guys for the review :) @ederign @jsoltes , could you pleas take a look?
Thanks

PR #429, opened 2018-08-06
[AF-1396] Asset editors don't resize with window size change

ederign:
The old method was overriding af default resize handler : https://github.com/kiegroup/appformer/blob/2b157a2ca6f8e4f2c2a9b02fc5cb731b1ddc64a4/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/Workbench.java#L244


tomasdavidorg:
Thanks. During the verification I found https://issues.jboss.org/browse/AF-1461

PR #448, opened 2018-09-05
JBPM-7665: Process Instances: page with vertical scroll bar is flicke…

nmirasch:
…ring after changing window width

Changed the pageSizeSelector style to dropup


nmirasch:
@cristianonicolai @bsremac  could you take a look?

nmirasch:
@cristianonicolai added the method to configure the pageSizeSelector dropup instead of setting that for all pagedTables

nmirasch:
@cristianonicolai  included the pageTable Column size change you proposed. Could you take a look?

PR #454, opened 2018-09-11
[AF-1502] New JGit Hook engine and also first hook (git fs external update - ssh)

ederign:


ederign:
Included in : https://github.com/kiegroup/appformer/pull/460#pullrequestreview-155274699

PR #456, opened 2018-09-12
latest Zanata translation changes in appformer from 2018-09-12

kiereleaseuser:


ederign:
@mbiarnes should we also have this on master?

PR #462, opened 2018-09-16
AF-1528: improvements and consitency on web tooling behavior if offline

porcelli:
FIxing some issues on offline behavior, the PRs that compose the changes are:

https://github.com/kiegroup/kie-soup/pull/59
https://github.com/kiegroup/appformer/pull/462
https://github.com/kiegroup/kie-wb-common/pull/2127
https://github.com/kiegroup/kie-wb-distributions/pull/814


uberfire-m2repo-editor/uberfire-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/FixNotMavenizedArtifactInfo.java
```diff
@@ -0,0 +1,61 @@
...
+            logger.error(e.getMessage(), e);
+        }
+        this.pomTemplate = pomTemplate;
+    }
+
+    public Properties getProperties(final String filePath) {
+        final Properties result = new Properties();
+        final String fullFileName = filePath.substring(filePath.lastIndexOf("/") + 1);
+        final String fileNamePart = fullFileName.substring(0, fullFileName.indexOf("-"));
+        final String fileVersion = fullFileName.substring(fullFileName.indexOf("-") + 1).replace(".jar", "");
```
jhrcek: This code for extracting file version looks very brittle. Even in this PR ensemble I see jar filenames like
`uberfire-m2repo-editor-backend-100-SNAPSHOT.jar` not to mention that you can have "1.3.2-SNAPSHOT" that also contains dash. You also have for example `ant-launcher` in one of the not-mavenized-artifacts.txt for which `fullFileName.substring(0, fullFileName.indexOf("-"));` will return `ant` -> will not find it in the properties for this artifacts. This should be made more robust and unit tested thoroughly. Please add cases like that to unit tests (fileNamePart containing multiple dashes, fileVersion containing SNAPSHOT etc).

uberfire-m2repo-editor/uberfire-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/FixNotMavenizedArtifactInfo.java
```diff
@@ -0,0 +1,61 @@
...
+            }
+        } catch (IOException e) {
+            logger.error(e.getMessage(), e);
+        }
+        this.pomTemplate = pomTemplate;
+    }
+
+    public Properties getProperties(final String filePath) {
+        final Properties result = new Properties();
+        final String fullFileName = filePath.substring(filePath.lastIndexOf("/") + 1);
```
jhrcek: Whenever you're doing `(last)IndexOf(String)` with just one character String, youshould replace it with overloaded version `(last)IndexOf(Char)`, which is faster. 3x in this PR

uberfire-m2repo-editor/uberfire-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/FixNotMavenizedArtifactInfo.java
```diff
@@ -0,0 +1,61 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(FixNotMavenizedArtifactInfo.class);
+
+    private final Properties notMavenizedArtifacts = new Properties();
+    private final String pomTemplate;
+
+    public FixNotMavenizedArtifactInfo() {
+        String pomTemplate = null;
+        try {
+            pomTemplate = new BufferedReader(new InputStreamReader(M2ServletContextListener.class.getResourceAsStream("template.pom")))
```
jhrcek: Any reason to split an then immediately join using newline? Why not just load the artifact as a string?

Also in this case (since we only have one fixed template which is relatively small) I'd simplify the code and include the template as a `private static final String` field into this class directly.

uberfire-m2repo-editor/uberfire-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/FixNotMavenizedArtifactInfo.java
```diff
@@ -0,0 +1,61 @@
...
+            if (isNotMavenizedArtifacts != null) {
+                notMavenizedArtifacts.load(isNotMavenizedArtifacts);
+            }
+        } catch (IOException e) {
+            logger.error(e.getMessage(), e);
+        }
+        this.pomTemplate = pomTemplate;
+    }
+
+    public Properties getProperties(final String filePath) {
```
jhrcek: I'm wondering why you're using Properties object to hold just Group / Artifact / Version and then calling `properties.getProperty(GROUP_ID)`. Wouldn't it be simpler to use some kind of `GAV` class instead?

porcelli: The reason for that is the consistency on callsite, as there it uses the "pom.properties" loaded into a Properties (what is correct in that context).

porcelli:
thank you very much @jhrcek for the feedback, just updated the code

uberfire-m2repo-editor/uberfire-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/FixNotMavenizedArtifactInfo.java
```diff
@@ -0,0 +1,90 @@
...
+import org.guvnor.common.services.project.model.GAV;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.guvnor.m2repo.backend.server.M2ServletContextListener.ARTIFACT_ID;
+import static org.guvnor.m2repo.backend.server.M2ServletContextListener.GROUP_ID;
+import static org.guvnor.m2repo.backend.server.M2ServletContextListener.VERSION;
+
+/**
+ * This class aim to provide maven meta information for artifacts (.jar) that don't
```
jhrcek: Thanks for the comment. I would have never figured out the intention of this just from the code.

uberfire-m2repo-editor/uberfire-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/FixNotMavenizedArtifactInfo.java
```diff
@@ -0,0 +1,90 @@
...
+
+    public FixNotMavenizedArtifactInfo() {
+        final Properties notMavenizedArtifactsProps = new Properties();
+        try {
+            final InputStream isNotMavenizedArtifacts = M2ServletContextListener.class.getResourceAsStream("/not-mavenized-artifacts.properties");
+            if (isNotMavenizedArtifacts != null) {
+                notMavenizedArtifactsProps.load(isNotMavenizedArtifacts);
+            }
+            for (Map.Entry<Object, Object> entry : notMavenizedArtifactsProps.entrySet()) {
+                notMavenizedArtifacts.put(entry.getKey().toString() + "-", entry.getValue().toString().split(":"));
```
jhrcek: Having property files with entries `artifactId=groupId:artifactId` in the properties file is redundant (and potential sourse of typo-bugs). It would have been better to have `artifactId=groupId`, in which case you wouldn't have to split and could read the artifact id from property value.

uberfire-m2repo-editor/uberfire-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/FixNotMavenizedArtifactInfo.java
```diff
@@ -0,0 +1,90 @@
...
+ * To build the GAV information out of a .jar file, this class uses as input a
+ * `not-mavenized-artifacts.properties` from classpath that provides the feedback needed
+ * to produce the required metadata. In case of an artifact is not listed in the input file,
+ * this class won't be able to provide the GAV.
+ */
+public class FixNotMavenizedArtifactInfo {
+
+    private static final Logger logger = LoggerFactory.getLogger(FixNotMavenizedArtifactInfo.class);
+
+    private final TreeMap<String, String[]> notMavenizedArtifacts = new TreeMap<>((o1, o2) -> {
```
jhrcek: It took me several minutes to understand this weird comparator. In the end I understood the intention, but without knowing the context, this might seem like a bug to sort long strings before shorter ones.

In this case comment and maybe more readable comparator using Java 8 apis would have been better. Also you don't need to do the final comparison of same-length strings as they wouldn't be equal anyway.

```java
    // Sort artifact names by length from longer to shorter to avoid false positives when shorter prefix matches too (e.g.  ant-launcher vs. ant)
    private static final Comparator<String> LONG_NAMES_BEFORE_SHORT_COMPARATOR = Comparator.comparing(String::length).reversed();
    private final TreeMap<String, String[]> notMavenizedArtifacts = new TreeMap<>(LONG_NAMES_BEFORE_SHORT_COMPARATOR);
```

uberfire-m2repo-editor/uberfire-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/FixNotMavenizedArtifactInfo.java
porcelli: Hi @jhrcek this is not exactly the case, for example we have this entry in the real-world properties file:
hibernate-jpa=org.hibernate.javax.persistence:hibernate-jpa-2.1-api

uberfire-m2repo-editor/uberfire-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/FixNotMavenizedArtifactInfo.java
porcelli: I could add the comment, and even make it a constant, but I don't think I could use the Java8 comparator, as the goal is not just sort by length, but also sort alphabetically

jhrcek: Looking at how the map keys are used in this class (`fullFileName.startsWith(entry.getKey())`) the order of prefixes with the same lenght should not matter. You can leave it as is now, but know (for the future) that java 8 apis are quite powerful. In particular the comparator that you have could have been made a bit more readable like so:

```java
    private static final Comparator<String> LONG_NAMES_BEFORE_SHORT_COMPARATOR = Comparator
            .comparing(String::length).reversed()
            .thenComparing(naturalOrder());
```

PR #463, opened 2018-09-17
DROOLS-2940 : It is not possible to run all scenarios in the old test scenarios.

Rikkola:
https://issues.jboss.org/browse/DROOLS-2940

Bundle:
https://github.com/kiegroup/appformer/pull/463
https://github.com/kiegroup/drools-wb/pull/935


Rikkola:
@kkufova @danielezonca Two fixes. One the wrong service was called from the "Run All". Two the event needs to fire only once for all the services and sum the results up.

kkufova:
I agree with @danielezonca about returning the `null`. I don't think changing the current approach in this small PR would significantly improve the overall problem in the codebase.

PR #464, opened 2018-09-17
AF-1528: incorporating suggestion from jan

porcelli:


uberfire-m2repo-editor/uberfire-m2repo-editor-backend/src/main/java/org/guvnor/m2repo/backend/server/FixNotMavenizedArtifactInfo.java
```diff
@@ -33,12 +35,12 @@
 
     private static final Logger logger = LoggerFactory.getLogger(FixNotMavenizedArtifactInfo.class);
 
-    private final TreeMap<String, String[]> notMavenizedArtifacts = new TreeMap<>((o1, o2) -> {
-        if (o1.length() != o2.length()) {
-            return o2.length() - o1.length();
-        }
-        return o1.compareTo(o2);
-    });
+    private static final Comparator<String> LONG_BEFORE_SHORT_AND_THEN_ALPHABETICALLY_COMPARATOR = Comparator
```
jhrcek: I still think it would work correctly even without `thenComparing(naturalOrder())`

And to make it perfect I'd add a comment explaining this weird comparator `\\sort artifact names from longer to shorter, to prevent false-positive matches with artifactIds whose names are prefixes of another artifactIds (e.g. ant vs. ant-launcher)`

porcelli: i think you're going too far on nit picking :)

jhrcek: Sorry, I'll relax a bit more in the future :-) It's just that lately I've see too many examples of people not paying enough attention to making the code as simple and as readable as possible.

porcelli:
@jhrcek sorry jan, but the natural order is mandatory

PR #469, opened 2018-09-19
AF-1585: avoid port failure if already in use, by automatically assigning a new available port

porcelli:
Avoid port failure if already in use, by automatically assigning a new available port in case of preferred port isn't available.
This change should improve our testing infrastructre, avoiding what's called `randon` test failures.

Docs: https://github.com/kiegroup/kie-docs/pull/1157

### DIFF
--- a/uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/daemon/common/PortUtil.java
+++ b/uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/daemon/common/PortUtil.java
@@ -1,0 +1,36 @@
+package org.uberfire.java.nio.fs.jgit.daemon.common;
+
+import java.io.IOException;
+import java.net.ServerSocket;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PortUtil {
+
+    private static final String ERROR_MESSAGE = "Error trying to find a free port.";
+    private static final Logger LOG = LoggerFactory.getLogger(PortUtil.class);
+
+    public static int validateOrGetNew(int preferredPort) {
+        if (preferredPort == 0 || isPortInUse(preferredPort)) {
+            if (preferredPort != 0) {
+                LOG.warn("Port {} already in use, system will automatically look for a new one.", preferredPort);
+            }
+            try (ServerSocket ss = new ServerSocket(0)) {
+                return ss.getLocalPort();
+            } catch (IOException e) {
+                LOG.error(ERROR_MESSAGE, e);
+                throw new RuntimeException(ERROR_MESSAGE);
+            }
+        }
+        return preferredPort;
+    }
+
+    private static boolean isPortInUse(int port) {
+        try (final ServerSocket ss = new ServerSocket(port)) {
+            return false;
+        } catch (Exception e) {
+        }
+        return true;
+    }
+}
--- a/uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/daemon/git/Daemon.java
+++ b/uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/daemon/git/Daemon.java
@@ -50,15 +50,20 @@ import org.eclipse.jgit.transport.resolver.RepositoryResolver;
 import org.eclipse.jgit.transport.resolver.ServiceNotAuthorizedException;
 import org.eclipse.jgit.transport.resolver.ServiceNotEnabledException;
 import org.eclipse.jgit.transport.resolver.UploadPackFactory;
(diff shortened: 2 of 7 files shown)


porcelli:

porcelli:
`Q: Is this a replacement for the mechanism we already have in tests here?`
A: THe currenct mechanism for tests are weak, the reason for that is the NIO2 will boot with JVM, and once initialized the JGitFileSystem will be active during the lifetime of the JVM, so this approach works for the first test executed for the first time, all other tests that are executed after, this code is useless.
The spread of such thing was needed, as we never know what is the first tests to be executed.
This PR has a proposal that builts-in the logic of free port, so it's similar, but it's built-in into core.

`Q: Also similar workaround was implemented by Max in kie-wb-common here. Would you suggest replacing all its usages with the mechanism you implemented in this PR?`
A: We can now just remove all freeport in the whole codebase, with the change proposed by this PR it's not needed anymore.

`Q: Could you please provide minimal code example in javadoc about how this class can be used to prevent port conflict in tests (how to properly configure filesystem in tests)?`
A: This is built-in internally to NIO2 Git backend, nobody needs (or should) interact with it.

`Q: Also there is a ton of places in tests that disable starting of niogit daemon using system property org.uberfire.nio.git.daemon.enabled=false. What would the general recommendation be when choosing between these 2 approaches to prevent port conflicts in tests?`
A: It's not a matter of choose as the free port will be built-in. With that said, majority of our tests don't need the deamon, so I think we could disablle in the parent pom in the root of the surefire (as I mentioned other day in IRC), and override only where would be needed (very few places, quite sure).

porcelli:
test failed not related
@jhrcek  are you ok with this?

domhanak:
@porcelli I am concerned about this being handled "behind the scene" because if user sets port for his daemon using our system property, I would expect that the port is going to  be used. When the port is not free I would expect an exception, because later, user might come back and use that property for some magic, but in reality the property is going to point to different port, then the daemon uses.
Let me know if my concern is valid.

jhrcek:
@domhanak good point. This might be even more relevant in case of docker image distributions, where you have to specify port mappings to expose the functionality outside the container.

porcelli:
@domhanak the proposed solution will log if the assigned port is not available, informing what's the new port.
in containerized setup, unlikely to happen, as the container developer has more control over port assignment (at least it's explicit).

Regardless, today we don't exactly fail, we just throw an exception and weird errors start to happen (port already in use, with issues coming from NIO2 - so not the easiest issue to spot).

porcelli:
@domhanak what's your suggestion?

domhanak:
@porcelli currently, if the ssh port or port for daemon is being occupied, the workbench will not deploy at all.  I will double check it and let you know, but this is what I experienced recently.

Imho, we should fix the tests selecting  correct unoccupied port and clearing the port after. Changing this for production environment could bring some UX inconsitencies and confusion for users. 

For example if Admin sets a port for git.deamon and makes a mistake,  he will not know something is wrong and everything will function as expected, but what is in the property will differ with what is actually in use. I doubt he will check logs and look for for some message regarding this. Currently the workbench fails to deploy so and logs show  "Port is already in use" exceptions.

Also I do not see the system property being updated once we find a different unoccupied port. What if someone tries to use this property but the daemon is listening on different port then the property specifies?

domhanak:
@porcelli My suggestion is to make the findFreePort method more robust and to clear the port found after the tests are done. But this needs more discussion, but I would aim towards fixing tests rather then changing this internally.

porcelli:
I disagree @domhanak - the daemon's don't affect directly BC behavior, have a new port assigned (and properly logged) is just a matter of remove potential issues for getting started experience.

The freeport solution for tests are not possible, as once the JGitFileSystemProvider is initialized you can't change it again - and this is by design (NIO2).

The best we can do is also document that a new port is assigned if the one assigned is already in use. Other than that I can't see any issue

ederign:
+1 for @porcelli's solution.

jhrcek:
I tend to agree with @domhanak. The issue I see with this approach is: 
1. user requests specific port by setting system property `org.uberfire.nio.git.daemon.port=1234`. 
2. When this port is occupied, the app finds another one (and it is logged) but the user still lives in the illusion that it's using port 1234, because he doesn't check logs.

I'm totally fine with finding free port if user doesn't specify it (which will be the case in tests after we remove all the workarounds we have in place now). But I'd prefer "fail fast" behavior if user specifies the port using system property. @porcelli @ederign Let's discuss this in person before we merge this.

domhanak:
+1 for properly documenting this, please link PR to documentation repo or JIRA with blocker.
I would still love to see fail-fast approach we currently have in production environment, would it be possible?
In the meantime, please check linked PR that should make the findFreePort more robust as an alternative.

porcelli:
@domhanak no it's not possible still have failfast as today. it's one way or another.

as I already mentioned too, it's not possible to make findFreePort test better

jhrcek:
We discussed this with @domhanak and we agreed to merge this as is, but only after this behavior is properly documented. @porcelli could you please document this and create a PR to kie-docs?

ederign:
@porcelli should I wait for the docs to merge this?

porcelli:
yes, but I'm out for the week; will provide it first thing next week

PR #480, opened 2018-09-24
JBPM-7696 - Add 50 and 100 options to paged table

cristianonicolai:


cristianonicolai:
Assembly:
- https://github.com/kiegroup/appformer/pull/480
- https://github.com/kiegroup/kie-wb-common/pull/2082
- https://github.com/kiegroup/jbpm-wb/pull/1192

cristianonicolai:
@porcelli @ederign with these new options ( 50 and 100 ),  I'm thinking if still makes sense for us to keep 5 and 15, would you have any restrictions to keep only ( 10, 20, 50 and 100 )?

bsremac:
@cristianonicolai +1 for removing 5 and 15. I see no value in such a fine granularity for the type of data that is being listed.

manstis:
@bsremac @cristianonicolai re: "...for the type of data that is being listed" don't forget the generic `PagedTable` is not just used by the Task List screen.. that said 5 and 15 should _probably_ fit comfortably into our existing screens so pages of these sizes is possibly more cumbersome to users than should those sizes "in full".

cristianonicolai:
@manstis sure we need to consider the wider adoption, I believe 10 is the default and will still be there. In terms of usage, I don't see much value in such small granularity ( 5 and 15 when we already have 10 and 20). Do you have any example/screen that you believe these options would make a difference for users?

jhrcek:
@cristianonicolai just heads up: there's still this unresolved issue related to choosing larger page size when using dashbuilder table displayer https://issues.jboss.org/browse/RHPAM-946

manstis:
@cristianonicolai No, I don't have any examples where removing 5 and 15 makes any difference.. I was just reminding @bsremac that _some_ components are used else where and therefore to be mindful of changes for a specific usage.

cristianonicolai:
To make these changes independent, I created another PR to handle removing the 5 and 15 options from the pagedtable, see https://github.com/kiegroup/appformer/pull/486

PR #484, opened 2018-09-26
JBPM-7732 - Review pagination logic to allow showing paging controls …

cristianonicolai:
…without total number of items

### DIFF
--- a/dashbuilder/dashbuilder-client/dashbuilder-renderers/dashbuilder-renderer-default/src/main/java/org/dashbuilder/renderer/client/table/TableDisplayer.java
+++ b/dashbuilder/dashbuilder-client/dashbuilder-renderers/dashbuilder-renderer-default/src/main/java/org/dashbuilder/renderer/client/table/TableDisplayer.java
@@ -18,7 +18,6 @@ package org.dashbuilder.renderer.client.table;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
-
 import javax.enterprise.context.Dependent;
 import javax.inject.Inject;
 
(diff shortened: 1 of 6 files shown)


cristianonicolai:
Assembly:
- https://github.com/kiegroup/jbpm-wb/pull/1207
- https://github.com/kiegroup/appformer/pull/484

cristianonicolai:
@bsremac testing this one needs to check not only the process and task dashboard tables but also the data set creation area. In there you can explore creating a data set for a kie server and a regular database table. They both use the same tabledisplayer component in the preview but the difference is that for the kie server one we don't have the row count to calculate the number of pages.

cristianonicolai:
@dgutierr would be nice to have your feedback on this as well.

dgutierr:
@cristianonicolai As far as I understand this is an improvement to support the kie-server datasets not returning the "totalRows" field, which the table displayer expects in order to enable pagination capabilities. If that's the case then your changes look good.

There might be an edge case though, what if the total number of rows (I mean the existing rows in the backend not the dataset rows returned in a kie-server's call) matches the page size? When the user clicks on next page, does the table show empty?

cristianonicolai:
@dgutierr thanks, yes this is to address this limitation in the kie server queries, we explicitly avoid the extra query in there for doing total counts.
True, we don't know if you hit the total number of rows or just for the page. In that case, the user will be able to click next and will get an empty result page. That is how the manage section works now and I tried to have the same behavior in here.

dgutierr:
@cristianonicolai ok. It's a minor issue though. Tests passed btw. Looks good to me :-)

ederign:
@cristianonicolai feel free to merge this when you have a position about https://github.com/kiegroup/jbpm-wb/pull/1207

PR #485, opened 2018-09-26
DROOLS-2922: CSS: Pop-over menu for DMN decision tables

manstis:
See https://issues.jboss.org/browse/DROOLS-2922


manstis:
Part of an ensemble:
- https://github.com/kiegroup/appformer/pull/485
- https://github.com/kiegroup/kie-wb-common/pull/2148

manstis:
This PR fixes an issue for nested grids (only used in the DMN Editor) when the "relative position" of a mouse event on a nested grid did not consider the grids absolute position (Lienzo's `getComputedLocation()` method walks parents to determine the absolute position relative to the root Layer).

PR #493, opened 2018-10-11
AF-1512: Add a preferences option to select Business Central Profiles/Entitlements

jesuino:
Supporting COMBO and adding an utility for testing ParametrizedCombo


uberfire-extensions/uberfire-preferences-ui-client/src/main/java/org/uberfire/ext/preferences/client/central/form/DefaultPreferenceForm.java
```diff
@@ -113,11 +118,21 @@ private void createPropertiesEditorCategory() {
             final PropertyEditorFieldInfo fieldInfo = createFieldInfo(propertyName,
                                                                       propertyType,
                                                                       propertyValue);
-
+            if(property.getValue() == PropertyFormType.COMBO) {
```
paulovmr: Missing space after `if`. There are other cases in this PR.

paulovmr: I think it is best to move this logic into the `createFieldInfo` method and add a test for this case.

jesuino: Hello @paulovmr thanks for the feedback. I fixed the ifs and moved the logic into `createFieldInfo` method. For testing would you suggest create a property of enum type in `PortablePreferenceMock` or create a new preference? If I have to modify `PortablePreferenceMock` would you kindly suggest the preferable way to re-generate `PortablePreferenceMockBeanGeneratedImpl` and `PortablePreferenceMockPortableGeneratedImpl`? Thanks!

jesuino:

jhrcek:
@jesuino just a general tip around handling multi-repo PRs for next time :smile: 
For example I wasn't aware that these other PRs existed.

Once you have a complete list of PRs, it is a good practice to 
- Collect a list of the PRs
- sort them in order of "X needs to be build before Y" (this order can be found e.g. [here](https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/script/repository-list.txt))
- Paste that list into the description of all the PRs and keep it up-to date

This makes navigation between PRs much easier and also people building it locally don't have to lookup the correct order if things change

PR #501, opened 2018-10-18
upgraded org.apache.lucene because of github security vulnebaribility

mbiarnes:


ederign:
@adrielparedes can you please take a look at this?

adrielparedes:

adrielparedes:
@ederign checking, I will wait for full downstream build because it could break ISPN and Elasticsearch in serveral places

adrielparedes:
@ederign @mbiarnes this PR is not enough, ES is broken because of Lucene versions, and I think ES 5.6 uses Lucene 6, so IDK if that upgrade could be possible

adrielparedes:
@mbiarnes @ederign I think we should discard this PR and plan an upgrade if it is necessary

ederign:
@adrielparedes  +1 
@mbiarnes thoughts?

mbiarnes:
@ederign @tiagobento  We use org.apache.lucene 6.6.1, JBoss EAP 7.2  5.5.5.
So for aligning nothing has been done as we can't downgrade. 
This version is required because of automatic GitHub security vulnerability, it is proposed to update to 7.1.0 version. Since all tests and downstreams etc. are red we should close this PR.

PR #511, opened 2018-10-31
 AF-1669: Support slashes in branch names

paulovmr:
Part of an ensemble:
* https://github.com/kiegroup/appformer/pull/511
* https://github.com/kiegroup/kie-wb-common/pull/2228

### DIFF
--- a/uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/spaces/SpacesAPIImpl.java
+++ b/uberfire-backend/uberfire-backend-server/src/main/java/org/uberfire/backend/server/spaces/SpacesAPIImpl.java
@@ -30,7 +30,7 @@ import org.uberfire.spaces.SpacesAPI;
 @Startup(StartupType.BOOTSTRAP)
 public class SpacesAPIImpl implements SpacesAPI {
 
-    private static final String PATH_PATTERN = "^[A-Za-z]+://([^/]+)/.*";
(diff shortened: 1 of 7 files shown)


paulovmr:
@barboras7 @ederign @porcelli Hi, if you could take a look, please. Thanks!

paulovmr:
jenkins retest this

barboras7:

barboras7:
@paulovmr Hi Paulo, sorry for the delay. The code looks good, but I found some issue regarding the allowed names for the branch:
- '---' is a valid branch name (and it shouldn't be); this also applies to '///' branch name 
- 'myBranch/' uses only allowed characters but the error only lists not allowed characters (it should probably say something like 'the branch name cannot begin/end with .... characters')
- 'abc$def' - the $ sign causes an error in the business central (see attached screenshot)

screenshot from 2018-11-08 09-40-44

paulovmr:
@barboras7 Good catches, thanks!

 * I fixed it for the ´---´ name (generalizing, every name starting with ´-´). The validation for ´///´ was already working here. 
 * Regarding the name ending with ´/´, the validation message says ´See the man page for `git check-ref-format` for more details.´ to cover this. The message was agreed with UX team to keep it simple, because there are a lot of rules for branch naming.
 * I fixed the problem for names containing ´$´.

paulovmr:

paulovmr:

PR #515, opened 2018-11-05
JBPM-7902 - Allow configuration to avoid invalidating session

cristianonicolai:


uberfire-extensions/uberfire-security/uberfire-servlet-security/src/main/java/org/uberfire/ext/security/server/BasicAuthSecurityFilter.java
```diff
@@ -83,7 +90,7 @@ public void doFilter(final ServletRequest _request,
         } finally {
             // invalidate session only when it did not exists before this request
             // and was created as part of this request
```
MarianMacik: This comment no longer applies probably? Now it has to be specifically defined to be invalidated, right?

uberfire-extensions/uberfire-security/uberfire-servlet-security/src/main/java/org/uberfire/ext/security/server/BasicAuthSecurityFilter.java
```diff
@@ -83,7 +90,7 @@ public void doFilter(final ServletRequest _request,
...
-            if (session == null) {
+            if (session == null && invalidate) {
```
ederign: @cristianonicolai I'm not sure about the old uses cases, but I wonder if there is no use case to invalidate==false and session == null (the old ones).

But I think invalidate default true cover this.

ederign:
@cristianonicolai what is the best qe for this?

cristianonicolai:
@ederign at this point would be @MarianMacik

uberfire-extensions/uberfire-security/uberfire-servlet-security/src/main/java/org/uberfire/ext/security/server/BasicAuthSecurityFilter.java
cristianonicolai: @ederign true, old cases still covered by the fact that the default still to invalidate. Only when the filter is specifically parametrized to avoid that ( websocket case ) it will be avoided.

uberfire-extensions/uberfire-security/uberfire-servlet-security/src/main/java/org/uberfire/ext/security/server/BasicAuthSecurityFilter.java
cristianonicolai: @MarianMacik correct, I adjusted the comment to better reflect that.

MarianMacik:
@ederign @cristianonicolai are we ready to merge here?

MarianMacik:
After discussion with @cristianonicolai it is ready.

PR #526, opened 2018-11-13
[DROOLS-3074] Implement placeholder on body cells

gitgabrio:
@manstis @Rikkola 

Scope of the present PR is to move inside appformer the "placeholder" logic for body grid cell


uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/renderers/columns/impl/IntegerColumnRenderer.java
```diff
@@ -27,20 +27,29 @@
...
-        t.setListening(false);
-        t.setX(context.getCellWidth() / 2);
-        t.setY(context.getCellHeight() / 2);
-        g.add(t);
-        return g;
+        Text text;
+        String value;
+
+        // Show placeholder only if the following conditions are met
+        if (cell.getValue() != null && cell.getValue().getValue() == null && cell.getValue().getPlaceHolder() != null) {
```
Rikkola: I would add a method here ( and in `StringColumnRenderer.java` ) that removes the comment above. Something like:
`if(isCellEmpty(cell))`
Just to remove the comment above and mention why the placeholder used.

And also test this if so that the next guy does not come and break your code.

ederign:
@gitgabrio @manstis we reverted this because it's not green  here https://github.com/kiegroup/drools-wb/pull/994 and this breaks the build.

Also, for all PR on appformer PR's is mandatory to always include a review from QE.

PR #533, opened 2018-11-16
Remove slf4j-ext which is not user anywhere

jhrcek:
You can check for yourself by running this in the appformer repo:
`grep 'org.slf4j.*' -r --include=*.java -o -h | sort | uniq`

You'll see that we're only ever using from slf4j is Logger and LoggerFactory that come from slf4j-api
If you're curious which functionality slf4j-ext (which we don't use) provides, you can check its [website](https://www.slf4j.org/extensions.html).

### DIFF
--- a/dashbuilder/dashbuilder-distros/pom.xml
+++ b/dashbuilder/dashbuilder-distros/pom.xml
@@ -120,11 +120,6 @@
     </dependency>
 
     <dependency>
-      <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-ext</artifactId>
-    </dependency>
-
-    <dependency>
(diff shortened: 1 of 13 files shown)


ederign:

jhrcek:

jhrcek:
Infrastructure problem during build:
`Caused by: java.io.IOException: Backing channel 'kie-ci-rhos-rhel7-xlarge-xmem-114589' is disconnected.`

jhrcek:
@ederign There are 9 failures, 3 of those are on master right now, the rest is kie-server integration test failures. So nothing related to this PR. I think we should merge this.

PR #537, opened 2018-11-19
[AF-1716] Branches should be copy only in the same FileSystem

ederign:
This bug was raised by https://github.com/kiegroup/appformer/commit/a7a057fdd9dff56e3633d87a1c00daa921f36829


ederign:

PR #541, opened 2018-11-20
AF-1452 Backend of Dynamic POM Dependencies framework

desmax74:
@ederign @pefernan @domhanak


desmax74:
jenkins retest this

desmax74:
jenkins retest this

uberfire-structure/uberfire-structure-api/src/main/java/org/guvnor/structure/pom/DependencyType.java
```diff
@@ -18,5 +18,8 @@
  * Types managed by the Pom's event
  */
 public enum DependencyType {
-    JPA
+    JPA,
+    TEST,
+    VALIDATION,
+    SCENARIO_RUNTIME
```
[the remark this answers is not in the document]
desmax74: These are the types/alias available in the UI, they maps the list of deps mapped in the DependencyTypesMapper.json

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/pom/PomEditorDefault.java
```diff
@@ -122,14 +158,134 @@ public boolean addDependencies(List<DynamicPomDependency> deps,
...
+            List<Dependency> depsFromPom = model.getDependencies();
+            if (removeDynamicDep(depsFromPom,
+                                 dep)) {
+                model.setDependencies(depsFromPom);
+                writePOMModelOnFS(filePath,
+                                  model);
+                result = true;
+            }
+        } catch (Exception ex) {
+            logger.error(ex.getMessage(),
```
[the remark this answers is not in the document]
desmax74: Ok

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/pom/PomEditorDefault.java
```diff
@@ -122,14 +158,134 @@ public boolean addDependencies(List<DynamicPomDependency> deps,
...
+            Map<String, String> dynamicKeys = getKeysFromDynamicDependencies(deps);
+            if (removeDynamicDeps(depsFromPom,
+                                  dynamicKeys)) {
+                model.setDependencies(depsFromPom);
+                result = true;
+                writePOMModelOnFS(filePath,
+                                  model);
+            }
+        } catch (Exception ex) {
+            logger.error(ex.getMessage(),
```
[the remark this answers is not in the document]
desmax74: ok

uberfire-structure/uberfire-structure-api/src/main/java/org/guvnor/structure/pom/DependencyType.java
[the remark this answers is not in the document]
domhanak: hmm, I am not sure I understand this correctly, what determines a type of a dependency? How can we say that this dependency is JPA type, this is TEST type etc?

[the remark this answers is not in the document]
desmax74: Is choosed by the user checking the checkbox on the UI, currently only persistable/JPA is available

uberfire-structure/uberfire-structure-backend/src/main/filtered-resources/DependencyTypesMapper.json
```diff
@@ -0,0 +1,38 @@
...
+  "internalArtifacts":[
+    {"groupId":"org.kie", "artifactId":"kie-internal", "scope":"provided"},
+    {"groupId":"org.optaplanner", "artifactId":"optaplanner-core", "scope":"provided"},
+    {"groupId":"org.dashbuilder", "artifactId":"kie-internal", "scope":"provided"}
+  ],
+  "dependencies":[
+    {
+      "type": "JPA",
+      "deps": [
+        {"groupId": "org.hibernate.javax.persistence", "artifactId": "hibernate-jpa-2.1-api", "version": "${version.org.hibernate.javax.persistence.hibernate-jpa-2.1-api}", "scope": "compile"}
```
domhanak: @desmax74 Please reach out to @mbiarnes  he is updating hibernate on master so it would be nice to align it here too

uberfire-structure/uberfire-structure-backend/src/test/java/org/guvnor/structure/backend/pom/PomEditorDefaultTest.java
```diff
@@ -113,112 +186,131 @@ public void addDepTest() {
...
-        List<DynamicPomDependency> deps = Arrays.asList(dep);
-        boolean result = editor.addDependencies(deps,
+        boolean result = editor.addDependencies(EnumSet.of(DependencyType.JPA),
                                                 PathFactory.newPath(tmp.toAbsolutePath().toString() + File.separator + POM,
                                                                     tmp.toUri().toString() + File.separator + POM));
         assertThat(result).isTrue();
     }
 
     @Test
     public void addDuplicatedDepTest() {
```
domhanak: @desmax74 Could you please add test where we try to add duplicate `junit` dependency but with different version.
E.g. there is a `junit:4.12` in pom but I want to add `junit:5.0`

uberfire-structure/uberfire-structure-api/src/main/java/org/guvnor/structure/pom/DependencyType.java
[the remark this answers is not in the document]
domhanak: Oh, okay. Thanks for the explanation. :) Are we covering additions via UI only?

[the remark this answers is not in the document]
desmax74: This is the automatic way, check a "type" in the UI and we will add the comphrensive list of dependencies needed is the specific editor, every editor present only the types suited for the specific editor

uberfire-structure/uberfire-structure-backend/src/test/java/org/guvnor/structure/backend/pom/PomEditorDefaultTest.java
domhanak: **User Scenarios**

- User is using `central`. He/she already have a project that runs smoothly, but user adds another persistable data-object. Suddenly his/hers project is not building anymore. If he/she checks pom.xml for JPA version, it is different than before (it is now changed to older version). Thing is that user decided to use newer `hibernate` version because he/she needs to, but it is overwritten when he/she creates new objects even if user does not want to.
> I suggest prompting user if he wants us to update the version in this case. So instead of overwriting the version we should let user make the decision and possibly remember his/hers decision

- User is upgrading from X.0 to X.1 version. His/hers projects (couple of hundreds) use `JPA` dependencies that have version older than our version in business-central. 
Do we update them or not? If we do, how do we update them? 
We might also break users projects without user explicitly knowing it until he tries to build them.

 @desmax74 @ederign thoughts? How should we handle these cases?

uberfire-structure/uberfire-structure-backend/src/main/filtered-resources/DependencyTypesMapper.json
domhanak: @desmax74 I think that new version has different groupId

kiereleaseuser2:
Build finished. 5034 tests run, 11 skipped, 0 failed.

uberfire-structure/uberfire-structure-backend/src/test/java/org/guvnor/structure/backend/pom/PomEditorDefaultTest.java
ederign: Hi @domhanak,

I think the mismatch here is between 'regular java projects' and 'kie java projects'.

Our projects need to have a straight alignment with our versions because in the end, they will run in kie-server.

So, "user decided to use newer hibernate version because he/she needs to"
"I want to add junit:5.0"

He can't. His version needs to be aligned with ours.

"User is upgrading from X.0 to X.1 version"

For this, we can discuss in the future a migration tool.


My call would be to at least at first, manage all pom dependencies and update them if necessary.

@porcelli wdyt?

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/pom/ConfigurationMap.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+package org.guvnor.structure.backend.pom;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import org.guvnor.structure.pom.DependencyType;
+import org.guvnor.structure.pom.DynamicPomDependency;
+/**
+ * Configuration loaded form JsonFile
```
ederign: from

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/pom/DependencyTypesMapper.java
```diff
@@ -26,18 +28,30 @@
...
+    public Set<DynamicPomDependency> getInternalArtifacts() {
+        return conf.getInternalArtifacts();
     }
 
-    public List<DynamicPomDependency> getDependencies(DependencyType key) {
-        return mapping.get(key);
+    public List<DynamicPomDependency> getDependencies(Set<DependencyType> dependencyTypes) {
+        List<DynamicPomDependency> result = new ArrayList<>();
+        for (DependencyType type : dependencyTypes) {
+            result.addAll(conf.getMapping().get(type));
```
ederign: It's not better and safe check if there is a configuration for that type on the JSON?

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/pom/PomEditorDefault.java
```diff
@@ -53,58 +62,85 @@ public boolean addDependency(DynamicPomDependency dep,
...
             Model model = getPOMModel(filePath);
-            Map<String, String> keys = getKeysFromDeps(model.getDependencies());
+            List<Dependency> depsFromPom = model.getDependencies();
+            Map<String, String> keys = getKeysFromDependencies(model.getDependencies());
 
             for (DynamicPomDependency dep : deps) {
-                String keyDep = getKeyFromDep(dep);
-                if (!keys.containsKey(keyDep)) {
-                    Dependency pomDep = getMavenDependency(dep);
+                boolean internalArtifact = isInternalArtifact(dep);
```
ederign: also this code looks like a duplication of addDependency right? Could you please do a refactoring?

uberfire-structure/uberfire-structure-api/src/main/java/org/guvnor/structure/pom/DependencyType.java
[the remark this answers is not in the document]
porcelli: I think we should try to make it a bit more open... enum will limit our needs or force us to change the enum every time we find a new use case

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/pom/PomEditorDefault.java
```diff
@@ -122,14 +158,139 @@ public boolean addDependencies(List<DynamicPomDependency> deps,
...
+                         dep.getGroupID(),
+                         ex.getMessage(),
+                         ex);
+            result = false;
+        }
+        return result;
+    }
+
+    @Override
+    public boolean removeDependencies(List<DynamicPomDependency> deps,
```
porcelli: I think we should not even consider remove dependency operation

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/pom/PomJsonReaderDefault.java
```diff
@@ -32,9 +34,11 @@
 import org.uberfire.java.nio.file.Files;
 import org.uberfire.java.nio.file.Paths;
 
-public class PomJsonReaderDefault implements PomJsonReader {
+public class PomJsonReaderDefault implements PomJsonReader{
```
porcelli: please follow kie group format rules

uberfire-structure/uberfire-structure-api/src/main/java/org/guvnor/structure/pom/DependencyType.java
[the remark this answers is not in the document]
desmax74: @porcelli  Depends on how much work we want to add on every use case, now on new case we add the dependencies on the json files and one line in the enum, if we want to use a dinamic configuration we have to write a meta configuration on every editor, scan on every start and provide a registry, is it this your idea or is different ?

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/pom/ConfigurationMap.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+public class ConfigurationMap {
+    private Map<DependencyType, List<DynamicPomDependency>> mapping;
+    private Set<DynamicPomDependency> internalArtifacts;
+    private String kieVersion;
+    public ConfigurationMap(Map<DependencyType, List<DynamicPomDependency>> mapping,
```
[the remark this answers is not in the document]
desmax74: Fixed

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/pom/DependencyTypesMapper.java
desmax74: Added a check about the presence

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/pom/PomEditorDefault.java
```diff
@@ -53,58 +62,85 @@ public boolean addDependency(DynamicPomDependency dep,
             return false;
         }
 
+        boolean result = false;
```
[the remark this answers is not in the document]
desmax74: Ok

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/pom/PomEditorDefault.java
```diff
@@ -53,58 +62,85 @@ public boolean addDependency(DynamicPomDependency dep,
...
                                    Path pomPath) {
+        List<DynamicPomDependency> deps = new ArrayList<>();
+        for (DependencyType dep : dependencyTypes) {
+            deps.addAll(mapper.getDependencies(EnumSet.of(dep)));
+        }
+
         if (deps.isEmpty()) {
             return false;
         }
         boolean result = false;
```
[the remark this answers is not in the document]
desmax74: Ok

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/pom/PomEditorDefault.java
```diff
@@ -122,14 +158,139 @@ public boolean addDependencies(List<DynamicPomDependency> deps,
         return result;
     }
 
-    private String getKeyFromDep(DynamicPomDependency dep) {
+    @Override
+    public boolean removeDependency(DynamicPomDependency dep,
+                                    Path pomPath) {
+
+        boolean result = false;
```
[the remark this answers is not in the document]
desmax74: Ok

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/pom/PomEditorDefault.java
desmax74: This is here because we have a maven goal to check if we have unused dependencies

uberfire-structure/uberfire-structure-api/src/main/java/org/guvnor/structure/pom/DependencyType.java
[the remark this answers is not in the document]
porcelli: It's not about the amount of work, it's about code architecture.. and this option to use enum looks very fragile to me.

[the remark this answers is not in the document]
desmax74: @porcelli changes applied

kiereleaseuser2:
Build finished. 4566 tests run, 10 skipped, 0 failed.

uberfire-structure/uberfire-structure-api/src/main/java/org/guvnor/structure/pom/DependencyType.java
```diff
@@ -17,6 +17,7 @@
 /***
  * Types managed by the Pom's event
  */
-public enum DependencyType {
-    JPA
+public interface DependencyType {
```
jhrcek: @desmax74 @porcelli @ederign  if for extensibility reasons we're switching to this inheritance-based solution for dependency types (instead of ENUM based solution), does it make sense to keep the "data" associated with each dependency type in one central `DependencyTypesMapper.json` file?
Wouldn't it be better to just have `DependencyType` have more methods, like `Set<GAV> getDependencies()` etc?
In that case we could get rid of all the code loading those data from json and mapping etc and the information for each dependency type would be more "centralized". What do you think?

porcelli: @jhrcek you're absolutely right! the only challenge is the V (version) of the GAV, that should be collected from our existing POM variables

desmax74: @jhrcek @porcelli the informations loaded from the json and from the Pom for each dependency are already contained in the DynamicDependencyPom https://github.com/kiegroup/appformer/pull/489/files#diff-22d70676e71cb2474a5c70629db02303 
the DependencyType works just as key in the ConfigurationMap https://github.com/kiegroup/appformer/pull/541/files#diff-eb740a937fd6b8ef9c34163fbc462c84

porcelli: @desmax74 could you point the json config files? as @jhrcek mentioned, we likely could have the GA hard coded in Java classes directly, and just collecting the V (version) from a config file that ideally should be fed by POM vars

desmax74: @porcelli here the file with the placeholders https://github.com/kiegroup/appformer/blob/1f6b657d1ec89344bc14dcb95be4720505c3beb2/uberfire-structure/uberfire-structure-backend/src/main/filtered-resources/DependencyTypesMapper.json

desmax74: The files contains the common (except SCENARIO_RUNTIME) dependencies like internals, test and validation, every module own a dedicated file when will be ready the infrastructure. Using a placeholder inside class means write a class for every dependency needed by each editor, instead to write one line in the config, if I have understand correctly your proposal.

porcelli: every editor should provide an interface that would provide a list of GAVs dependencies, and in this file that a specific editor would need to provide you could have multiple placeholder inside.. but still one interface `per editor` or per component.

desmax74: @porcelli ok every impl type (like JPA https://github.com/kiegroup/appformer/blob/1f6b657d1ec89344bc14dcb95be4720505c3beb2/uberfire-structure/uberfire-structure-api/src/main/java/org/guvnor/structure/pom/types/JPADependencyType.java) will contains the dependencies informations

desmax74: @porcelli sound good now with the placeholder in the java templates replaced before the build ?

desmax74: It's fine but in this way the Config become a registry (is already present an addDependencies) for the others modules' dependencies, is this ok ?

desmax74: Done

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/pom/PomEditorDefault.java
desmax74: @porcelli cleanup deps https://github.com/kiegroup/kie-wb-common/pull/2184

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/pom/PomEditorDefault.java
```diff
@@ -53,58 +62,85 @@ public boolean addDependency(DynamicPomDependency dep,
...
-                List<Dependency> modelDeps = model.getDependencies();
-                for (Dependency modelDep : modelDeps) {
+                for (Dependency modelDep : depsFromPom) {
                     if (modelDep.getGroupId().equals(dep.getGroupID()) && modelDep.getArtifactId().equals(dep.getArtifactID())) {
-                        modelDep.setVersion(versionKey);
+                        if (internalArtifact) {
+                            modelDep.setVersion(getKieVersionFromPom(dep,
+                                                                     depsFromPom));
+                        } else {
+                            modelDep.setVersion(keys.get(keyDep));
```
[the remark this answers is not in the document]
desmax74: Changed

uberfire-structure/uberfire-structure-api/src/main/java/org/guvnor/structure/pom/DependencyType.java
desmax74: @porcelli is it ok with the latest changes ?

uberfire-structure/uberfire-structure-api/src/main/java/org/guvnor/structure/pom/DynamicPomDependency.java
```diff
@@ -47,6 +49,30 @@ public String getScope() {
...
+            return true;
+        }
+        if (!(o instanceof DynamicPomDependency)) {
+            return false;
+        }
+        DynamicPomDependency that = (DynamicPomDependency) o;
+        return getGroupID().equals(that.getGroupID()) &&
+                getArtifactID().equals(that.getArtifactID()) &&
+                getVersion().equals(that.getVersion()) &&
+                Objects.equals(getScope(),
```
porcelli: Not sure if Objects is avialble on client side

uberfire-structure/uberfire-structure-api/src/main/java/org/guvnor/structure/pom/DynamicPomDependency.java
```diff
@@ -47,6 +49,30 @@ public String getScope() {
...
+                               that.getScope());
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(getGroupID(),
```
porcelli: same here, please check of Objects type is available in client-side

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/pom/PomEditorDefault.java
porcelli: we agreed to not support remove dependencies, so I think this should be cleaned.
@ederign correcT?

uberfire-structure/uberfire-structure-backend/src/test/java/org/guvnor/structure/backend/pom/PomEditorDefaultTest.java
```diff
@@ -37,15 +42,22 @@
...
+        configurationMap = new DynamicDependencyTypeConfigurationMap();
+        editor = new PomEditorDefault(configurationMap);
         tmpRoot = Files.createTempDirectory("repo");
         tmp = TestUtil.createAndCopyToDirectory(tmpRoot,
                                                 "dummy",
                                                 "target/test-classes/dummy");
+        DependencyType depType = new FakeJPATypeDependency();
+        configurationMap.addDependencies(depType,
+                                         depType.getDependencies());
+        JPA_HIBERNATE_VERSION = configurationMap.getMapping().get(depType.getType()).get(0).getVersion();
```
porcelli: why hibernate test here?

uberfire-structure/uberfire-structure-backend/src/test/java/org/guvnor/structure/backend/pom/PomStructureEditorTest.java
```diff
@@ -33,26 +36,33 @@
...
     public void setUp() throws Exception {
         tmpRoot = Files.createTempDirectory("repo");
         tmp = TestUtil.createAndCopyToDirectory(tmpRoot,
                                                 "dummy",
                                                 "target/test-classes/dummy_empty_deps");
+        configurationMap = new DynamicDependencyTypeConfigurationMap();
+        DependencyType depType = new FakeJPATypeDependency();
+        configurationMap.addDependencies(depType,
+                                         depType.getDependencies());
+        JPA_HIBERNATE_VERSION = configurationMap.getMapping().get(depType.getType()).get(0).getVersion();
```
porcelli: why jpa here? if it's a `random` item.. please try not use the JPA to avoid confusion with real JPA dependency

desmax74: @porcelli Is a Fake dependency to test the read of the mapping. why Random ?

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/pom/PomEditorDefault.java
desmax74: @porcelli  @ederign I'll remove the removal

uberfire-structure/uberfire-structure-api/src/main/java/org/guvnor/structure/pom/DynamicPomDependency.java
desmax74: @porcelli I'll change with a more safe check

uberfire-structure/uberfire-structure-api/src/main/java/org/guvnor/structure/pom/DynamicPomDependency.java
desmax74: @porcelli I'll change with a more safe check

uberfire-structure/uberfire-structure-backend/src/test/java/org/guvnor/structure/backend/pom/PomEditorDefaultTest.java
desmax74: @porcelli Is a Fake dependency, I'll change the name

ederign:
I'm closing this PR due to inactivity. If necessary and I'll reopen it again, just let me know.

PR #544, opened 2018-11-22
[AF-1718] NullPointerException is thrown while creating multiple projects through curl command

elguardian:
Establishing order between workbench indexing and file system indexing


elguardian:
@Rikkola wdyt ? problem here is that they are colliding.

Rikkola:
Asking around for someone to test and review.

Rikkola:
Ok @tomasdavidorg volunteered.

ederign:

ederign:
Hello @elguardian could you please give me more details of the issue and also on your solution?

it's not clear what you mean by: 

"Establishing order between workbench indexing and file system indexing" 

"problem here is that they are colliding."

elguardian:
I am sequencing BatchIndex and the first time and setupWatchService is otherwise they clash in the beginning during fs creation as they manipulate the indexes at the same time.
The deletion event mentioned in the jira https://issues.jboss.org/browse/RHPAM-1713 is coming from the watch service.

porcelli:
if @tomasdavidorg can't see it on master anymore, we should avoid merge this

PR #551, opened 2018-11-28
[AF-1725] Make it possible to identify currently enabled business-central profile (for selenium testing)

ederign:
@jesuino @jhrcek does that works for you?

https://cl.ly/90d703c9cfea/%255B6fa93efe9abf842e880d3dfec28d4e8e%255D_Image%2525202018-11-27%252520at%2525207.22.09%252520PM.png

@jesuino please fire the event when the profile is selected/changed (also on startup)

@jhrcek $('.uf-workbench-layout') is a selector that will containg the css class.


jesuino:
@ederign there's one issue, it is not cleaning the previous added class, which means that if an user change class it will show two classes -> https://pasteboard.co/HPgsBda.png

However, for @jhrcek this may not be an issue, because when you start business-central it contains the current selected profile, the issue is only when user change class, it will fire an event and will accumulate profile classes.

jhrcek:
Didn't check this in the running workbench. I agree with William that It would be preferable if it worked even after switching profiles.

ederign:
jenkins retest this

PR #569, opened 2018-12-11
AF-1655: Make WB dashbuilder integration configurable to allow users select an offline alternative

jesuino:
Adding support to to the system property `org.dashbuilder.renderer.default`, which can have values c3 and gwtcharts on Business Central.

The following PRs are related to this:

https://github.com/kiegroup/kie-wb-distributions/pull/857
https://github.com/kiegroup/jbpm-wb/pull/1254

edited by eder:
Also related.
https://github.com/kiegroup/appformer/pull/584

### DIFF
--- a/dashbuilder/dashbuilder-backend/dashbuilder-services/src/main/java/org/dashbuilder/renderer/service/RendererSettingsImpl.java
+++ b/dashbuilder/dashbuilder-backend/dashbuilder-services/src/main/java/org/dashbuilder/renderer/service/RendererSettingsImpl.java
@@ -1,0 +1,19 @@
+package org.dashbuilder.renderer.service;
+
+
+import javax.enterprise.context.ApplicationScoped;
+
+import org.jboss.errai.bus.server.annotations.Service;
+
+@Service
+@ApplicationScoped
+public class RendererSettingsImpl implements RendererSettings {
+
+    String DEFAULT_RENDERER_PROPERTY = "org.dashbuilder.renderer.default";
+    
+    @Override
+    public String userDefaultRenderer() {
+        return System.getProperty(DEFAULT_RENDERER_PROPERTY);
+    }
+
+}
--- a/dashbuilder/dashbuilder-client/dashbuilder-displayer-client/src/main/java/org/dashbuilder/displayer/client/RendererManager.java
+++ b/dashbuilder/dashbuilder-client/dashbuilder-displayer-client/src/main/java/org/dashbuilder/displayer/client/RendererManager.java
@@ -32,7 +32,7 @@ import org.dashbuilder.displayer.DisplayerSettings;
 import org.dashbuilder.displayer.DisplayerSubType;
 import org.dashbuilder.displayer.DisplayerType;
 import org.dashbuilder.displayer.client.resources.i18n.CommonConstants;
-import org.dashbuilder.renderer.service.RendererSettings;
+import org.dashbuilder.renderer.service.RendererSettingsService;
 import org.jboss.errai.common.client.api.Caller;
 import org.jboss.errai.ioc.client.api.EntryPoint;
 import org.jboss.errai.ioc.client.container.SyncBeanDef;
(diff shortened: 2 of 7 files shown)


jesuino:
Captured a new bug with bubble chart on Process Instances dashboard -> 

https://issues.jboss.org/browse/AF-1749

jesuino:
@ederign @dgutierr @jhrcek Would you please take a look when you have a moment? There one issue I found: https://issues.jboss.org/browse/AF-1749

Besides that the integration seems to have worked, c3 can be used to replace Google and dashboards work offline now.

There are a few known limitations, which will be addressed in other JIRAs:

1) C3 does not support Meter, 3D Pie Chart and Maps types. There are plans to fix each not supported chart:

1.1: 3D Pie Chart will actually use the 2d Pie Chart, and it will be documented;
1.2: Map will use DataMaps -> https://issues.jboss.org/browse/AF-1663
1.3 Pendent: Select an API for Meter.


2) If you set C3 as the default you will still see maps and meter in the pallet of components, this will be handled here -> https://issues.jboss.org/browse/AF-1742
3) If an user added a type that is specific to Google to a page (e.g. map) and Google is not available an error is shown, we will improve that message -> https://issues.jboss.org/browse/AF-1742

There are two known issues:

1) Area Chart is not displayed: https://issues.jboss.org/browse/AF-1739
2) Bubble chart is throwing an error in certain conditions: https://issues.jboss.org/browse/AF-1749

kiereleaseuser2:
Build finished. 5050 tests run, 11 skipped, 0 failed.

dashbuilder/dashbuilder-backend/dashbuilder-services/src/main/java/org/dashbuilder/renderer/service/RendererSettingsImpl.java
```diff
@@ -0,0 +1,19 @@
...
+
+@Service
+@ApplicationScoped
+public class RendererSettingsImpl implements RendererSettings {
+
+    String DEFAULT_RENDERER_PROPERTY = "org.dashbuilder.renderer.default";
+    
+    @Override
+    public String userDefaultRenderer() {
+        return System.getProperty(DEFAULT_RENDERER_PROPERTY);
```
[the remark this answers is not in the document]
jesuino: Thanks for your review.

I was expecting it to indeed return null, I tried Optional, but Errai remote does not seem to work with Optional at the moment.  however, I can change to return some specific value, such as Google Chart UUID.

The issue with returning a default value or using an enum is that Renderers are decoupled from the main API, it is discovered using CDI so it give us the flexibility to plug/unplug renderers. So the renderer API should not know any specific renderer.

What you think about these two actions to avoid NPEs:

1) Return by default an empty String;
2) Comment the method saying that it returns an empty String if no value is set by the users.

Thanks!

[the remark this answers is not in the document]
jesuino: @jhrcek Please let me know if you think that returning an empty String is a good approach and I can modify it tomorrow.
I will work back on this PR once https://github.com/kiegroup/appformer/pull/573 is merged

Thanks!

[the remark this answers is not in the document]
jhrcek: Sorry, I missed your comment 2 days ago. Maybe I didn't receive email notification, because you didn't tag me in the comment?

> I tried Optional, but Errai remote does not seem to work with Optional at the moment.

Is that because Optional is not serializable? 

Empty String + nullcheck / check for empty string is fine with me.

dashbuilder/dashbuilder-client/dashbuilder-displayer-client/src/main/java/org/dashbuilder/displayer/client/RendererManager.java
```diff
@@ -56,15 +62,28 @@ public RendererManager(SyncBeanManager beanManager) {
...
+        rendererSettings.call((String defaultUUIDOp) -> lookupRenderers(defaultUUIDOp))
+                        .userDefaultRenderer();
+    }
+
+    private void lookupRenderers(String defaultUUID) {
         Collection<SyncBeanDef<RendererLibrary>> beanDefs = beanManager.lookupBeans(RendererLibrary.class);
+        if (defaultUUID != null && ! defaultUUID.isEmpty()) {
+            beanDefs.stream()
+                    .map(SyncBeanDef::getInstance)
+                    .filter(render -> render.getUUID().equals(defaultUUID))
```
jhrcek: I'm just wondering what will happen if NONE of the renderers will have uuid matching the user-selected
E.g what happens when user sets `-Dorg.dashbuilder.renderer.default=nonexistentUUID`?
Maybe that scenario is ok, but I'm not sure just from looking at the code..

jesuino: @jhrcek  updated with a new commit, would you please take a look? If user set a bad value for the system property it is simply ignored. I added tests for the changes as well.

jesuino: @jhrcek Regarding the issue you found, I could reproduce it and found a pattern, it only happens when there's charts with "no data". I will reproduce it on dashbuilder and let you know.

jesuino: Just for context, @jhrcek  tested this PR and saw the following error --> http://i.imgur.com/7BeOyKD.png

I fixed AF-1749 and the error seems to be gone. In fact, the error Jan saw is a generic error, which hides the root cause, and the root cause seems to be indeed AF-1749.

AF-1749 was fixed on the PR [1]. @jhrcek  Please let me know if you prefer to check  [1] and then come back to this PR or I can cherry pick the commit from [1] to this PR and you can test everything at once.

Thanks!

[1] https://github.com/kiegroup/appformer/pull/584

PR #554, opened 2018-11-28
[AF-1713]: Added clustered event when deleting project

adrielparedes:
@jhrcek Added Clustered to DeleteProjectEvent. Now, when a project is deleted all nodes of the cluster are informed. If they are in the same View and any node deletes the project, view is closed and all return to LibraryView.

PR #571, opened 2018-12-11
[AF-1744] File System Cache is not updated when project is deleted

ederign:
screenshot 2018-12-11 15 10 28

This clears the FS cache after a project deletion on another node.

Drools wb War to test it: https://www.dropbox.com/s/j6pd27uzfke1rjt/drools-wb-webapp.war?dl=0

@jhrcek and @adrielparedes it's important to keep the library state like the video (on space screen) to avoid other nonrelated issues (that will be addressed later).

This PR was built over https://github.com/kiegroup/appformer/pull/554

@jhrcek @adrielparedes what do you think of after approval merging this small chunks?

Like this one and https://github.com/kiegroup/appformer/pull/554 ?

### DIFF
--- a/uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/JGitFileSystemProvider.java
+++ b/uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/JGitFileSystemProvider.java
@@ -409,7 +409,7 @@ public class JGitFileSystemProvider implements SecuredFileSystemProvider, Dispos
         fsManager.getOpenFileSystems().forEach(JGitFileSystem::close);
         shutdownSSH();
         forceStopDaemon();
-        fsManager.clear();
+        fsManager.shutdown();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


jhrcek:
@ederign 
> what do you think of after approval merging this small chunks?

I agree. Let's start with merging https://github.com/kiegroup/appformer/pull/554 

I'm testing the changes using your war, but I'm running into some problems. Basically every time I delete a project I'm getting https://gist.github.com/jhrcek/5e2bff10923be4131a0a4262b16ba284

I tried rebuilding appformer from this PR + business-central but had the same issue. Not sure what's causing this (I double checked that my user has the rights to `org.uberfire.nio.git.dir` but no luck there. 

Are you using the [setup](https://docs.google.com/document/d/1RppqiCHINxVXSiY1xyD6G36S9HNb3y_DH4F54rQQbsI) from the beginning of the doc to "simulate" filesystem synchronization?

When I try WITHOUT nfs server (both business-central pointing to the same dir using  `org.uberfire.nio.git.dir` it works fine as you're showing in the gif.
But when filesystem synchronization is involved (see the doc mentioned above) it's throwing errors very often.

ederign:
@adrielparedes are we going to include this in your new PR? Should I close this?

jhrcek:
+1 for closing this PR and including this with Adriel's fix. As it is, this doesn't fix any error, but instead it introduces one new error, that appears every time when the project is deleted, as I mention in one of the previous comments above.

PR #582, opened 2018-12-17
JBPM-7855 - Stunner cannot export diagrams (after any exception on client)

tiagodolphine:
The problem was caused after any exception on client, the native libraries used on the export was not working anymore, seems that they were unloaded, getting NPE when trying to acess them on all the export formats.

**Solution:** 

Changed the native libs that use JsInterop setting the namespace **GLOBAL**:
 `@JsType(isNative = true, namespace = JsPackage.GLOBAL)`. 
Some adjustments were necessary on the `FileExportScriptInjector` to make these changes work.

Now if we get any exception on client, it is still possible to export Diagrams without any error, the native libraries are still loaded as expect.

@romartin @hasys

### DIFF
--- a/uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/exports/jso/FileExportScriptInjector.java
+++ b/uberfire-extensions/uberfire-commons-editor/uberfire-commons-editor-client/src/main/java/org/uberfire/ext/editor/commons/client/file/exports/jso/FileExportScriptInjector.java
@@ -35,8 +35,8 @@ import org.uberfire.ext.editor.commons.client.file.exports.FileExportResources;
 @ApplicationScoped
 public class FileExportScriptInjector {
 
-    public static final String NS = "org.uberfire.ext.editor.commons.client.file.exports.jso.";
     public static final String NS_SEPARATOR = ".";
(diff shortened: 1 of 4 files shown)


tiagodolphine:

tiagodolphine:
@hasys analyzing the code, if you take a look on the `FileExportProducer` this is responsible to call the 'FileExportScriptInjector.inject()' and these classes are singletons where the scripts are loaded on the `ScriptInjector.TOP_WINDOW` that is the top level `window` object on JS.  In this way, the scripts are loaded just once and shared on the WB runtime by all callers, this is my understanding.
This fix is just to keep the namespace as Global, because it seems this namespace scope was being lost after the exceptions don't know why exactly, and the lib were returning `undefined`. 
@romartin do you have more info regarding the scripts injection other than these I mentioned?

Thanks !

romartin:
@tiagodolphine @hasys nope, I don't have any other comment, as @tiagodolphine comments above the injection should only take place one, this fix just changes the namespace for the exported js object, so I don't see any issues here...

ederign:
@hasys  can you please take a look on this?

hasys:
Hi @tiagodolphine,

I was able to see an exception and still was able to download the process. Thank you!
works good, thank you!

tiagodolphine:
Thanks for the review @hasys .

@ederign PR approved ready to be merged.

PR #583, opened 2018-12-18
AF-1758: Use Patternfly C3 CSS classes instead injecting C3 CSS

jesuino:
Removed C3 default CSS and let patternfly C3 CSS take over.

### DIFF
--- a/dashbuilder/dashbuilder-client/dashbuilder-renderers/dashbuilder-renderer-c3/src/main/java/org/dashbuilder/renderer/c3/client/exports/ResourcesInjector.java
+++ b/dashbuilder/dashbuilder-client/dashbuilder-renderers/dashbuilder-renderer-c3/src/main/java/org/dashbuilder/renderer/c3/client/exports/ResourcesInjector.java
@@ -1,7 +1,5 @@
 package org.dashbuilder.renderer.c3.client.exports;
 
-import com.google.gwt.dom.client.StyleInjector;
-
 public class ResourcesInjector {
 
     static boolean injected;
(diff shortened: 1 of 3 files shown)


jesuino:
@ederign @jhrcek I think it is better to just remove C3 CSS and let patternfly act, we are avoiding duplication and it even looks better and consistent  -> https://issues.jboss.org/secure/attachment/12444863/patternflyc3.png

dashbuilder/dashbuilder-client/dashbuilder-renderers/dashbuilder-renderer-c3/src/main/java/org/dashbuilder/renderer/c3/client/exports/ResourcesInjector.java
```diff
@@ -14,8 +12,6 @@ public static void ensureInjected() {
     }
 
     private static void injectAllResources() {
-        String c3css = NativeLibraryResources.INSTANCE.c3css().getText();
```
jhrcek: I'm wondering if we can now also remove this c3css from [here](https://github.com/kiegroup/appformer/blob/b9f688c6a8fe64b719cf8f332412039d83e2afd3/dashbuilder/dashbuilder-client/dashbuilder-renderers/dashbuilder-renderer-c3/src/main/java/org/dashbuilder/renderer/c3/client/exports/NativeLibraryResources.java) since you removed the (only?) usage + the associate resource css file? I didn't find any other usage in kiegroup repos. This way nobody will have to wonder in the future "is anyone using this css file?"

jesuino: I completely removed it in my last commit, I kept in case patternfly is updated, but if that happens we can find this PR and bring back the c3 default CSS. Thanks for the suggestion @jhrcek

PR #584, opened 2018-12-18
AF-1749: C3 Bubble Chart does not filter data correctly and display an error

jesuino:
Fixing noData state. Now when there's no data the chart won't even attempt to render, instead it will show a no data info. This fixes the bubble chart issue.

### DIFF
--- a/dashbuilder/dashbuilder-client/dashbuilder-renderers/dashbuilder-renderer-c3/src/main/java/org/dashbuilder/renderer/c3/client/C3DisplayerView.java
+++ b/dashbuilder/dashbuilder-client/dashbuilder-renderers/dashbuilder-renderer-c3/src/main/java/org/dashbuilder/renderer/c3/client/C3DisplayerView.java
@@ -10,6 +10,7 @@ import org.jboss.errai.common.client.dom.HTMLElement;
 import org.jboss.errai.common.client.ui.ElementWrapperWidget;
 import org.gwtbootstrap3.client.ui.Label;
 
+import com.google.gwt.core.client.GWT;
 import com.google.gwt.user.client.ui.FlowPanel;
 import com.google.gwt.user.client.ui.HTML;
 import com.google.gwt.user.client.ui.Panel;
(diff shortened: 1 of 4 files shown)


jesuino:
@ederign @jhrcek Could you please review this? The steps to reproduce the issue are in the JIRA, it can be reproduced also on dashbuilder and I think this may be related to what is impeding the integration from work correct (AF-1655) or at least without this fix we won't be able to get AF-1655 merged.

jhrcek:
Just a note: during testing of this I found an old issue (unrelated to this PR - so it doesn't block merging this PR). Just putting it here cause it's related to bubble charts: https://issues.jboss.org/browse/RHPAM-1810

jesuino:
@jhrcek thank you again, Jan!

dashbuilder/dashbuilder-client/dashbuilder-renderers/dashbuilder-renderer-c3/src/main/java/org/dashbuilder/renderer/c3/client/C3DisplayerView.java
```diff
@@ -74,4 +76,22 @@ public void setBackgroundColor(String color) {
                           .setBackgroundColor(color);
     }
     
+    @Override
+    public void noData() {
+        FlowPanel noDataPanel = new FlowPanel();
```
ederign: Hi @jesuino could you please use GWT.create here?

ederign: https://stackoverflow.com/questions/2242399/why-use-gwt-create-instead-of-new

dashbuilder/dashbuilder-client/dashbuilder-renderers/dashbuilder-renderer-c3/src/main/java/org/dashbuilder/renderer/c3/client/C3DisplayerView.java
```diff
@@ -74,4 +76,22 @@ public void setBackgroundColor(String color) {
...
+        noDataPanel.setWidth(width + "px");
+        noDataPanel.setHeight(height + "px");
+        Label lblNoData = new Label(C3DisplayerConstants.INSTANCE.common_noData());
```
ederign: same here.

jesuino:
@ederign resolved conflicts and used GWT.create instead new, thanks!

PR #601, opened 2019-01-15
AF-1617: Dashboard content does not auto fit to the screen after window resize

jesuino:


jesuino:
@jhrcek @ederign Would you please check? 

This is how it looks like -> https://issues.jboss.org/secure/attachment/12445518/chart_auto_resize.ogv

dashbuilder/dashbuilder-client/dashbuilder-renderers/dashbuilder-renderer-c3/src/main/java/org/dashbuilder/renderer/c3/client/C3DisplayerView.java
```diff
@@ -95,4 +102,29 @@ public void setSize(int width, int height) {
         this.height = height;
     }
     
-}
+    public void setResizable(boolean resizable, int maxWidth, int maxHeight) {
+        if (resizable) {
```
ederign: I think we should throw a runtime exception if we try to setReziable in a not resizable view.

dashbuilder/dashbuilder-client/dashbuilder-renderers/dashbuilder-renderer-c3/src/main/java/org/dashbuilder/renderer/c3/mutationobserver/MutationObserverInitWrapper.java
```diff
@@ -0,0 +1,20 @@
...
+
+import elemental2.dom.MutationObserverInit;
+import jsinterop.annotations.JsOverlay;
+import jsinterop.annotations.JsPackage;
+import jsinterop.annotations.JsType;
+import jsinterop.base.Js;
+import jsinterop.base.JsPropertyMap;
+
+
+//This can be removed once Elemental is updated do RC1, where a create method will be available for MutationObserverInit
```
ederign: @tiagobento can we update our Elemental implementation for RC1?

tiagobento: @ederign I don't see any reason against it. We should open PRs on `errai`, `droolsjbpm-build-botstrap`, and `appformer` with the upgrade and wait for the full downstream build.. Since we were using a beta version of Elemental2, there might've been some breaking changes comparing to RC1.

dashbuilder/dashbuilder-client/dashbuilder-renderers/dashbuilder-renderer-c3/src/main/java/org/dashbuilder/renderer/c3/client/C3DisplayerView.java
jesuino: Hello,

Thanks for your comments.

> I think we should throw a runtime exception if we try to setReziable in a not resizable view.

This should never happen because each displayer must provide if it supports a certain property or not (e.g. resizable). If we don't say that certain view does not support the property then it will always be false, meaning that the setResizable method will never be called;

> The current setResize api feels strange.

I do agree with you, I was making some changes when it is not resizable (clean the width when it is not resizable), but it does not seem to be necessary. Will change the API.


> when the resize is disabled I would disable the input fields for max width / max height in the UI to avoid confusion about why they don't have any effect. Even better would be to hide the width/height when the resize checkbox is unchecked (but you can ignore this point if this would be hard to implement in the current displayer settings).
> I would rename the labels from width / height to max width / max height to avoid confusion like "Why do I have to set width if it's resizable?"

Although I agree with both improvements, currently it is indeed complicate to implement this on specific ticket IMO. It would run out of the main scope from the original JIRA. But I can create new JIRAs for future investigation.

dashbuilder/dashbuilder-client/dashbuilder-renderers/dashbuilder-renderer-c3/src/main/java/org/dashbuilder/renderer/c3/client/charts/bar/C3BarChartDisplayer.java
```diff
@@ -72,20 +73,9 @@ public DisplayerConstraints createDisplayerConstraints() {
...
-                .supportsAttribute(DisplayerAttributeGroupDef.FILTER_GROUP)
-                .supportsAttribute(DisplayerAttributeGroupDef.REFRESH_GROUP)
-                .supportsAttribute(DisplayerAttributeGroupDef.GENERAL_GROUP)
-                .supportsAttribute(DisplayerAttributeDef.CHART_WIDTH)
-                .supportsAttribute(DisplayerAttributeDef.CHART_HEIGHT)
-                .supportsAttribute(DisplayerAttributeDef.CHART_BGCOLOR)
-                .supportsAttribute(DisplayerAttributeGroupDef.CHART_MARGIN_GROUP)
-                .supportsAttribute(DisplayerAttributeGroupDef.CHART_LEGEND_GROUP)
-                .supportsAttribute(DisplayerAttributeGroupDef.AXIS_GROUP);
+        return new CommonC3DisplayerConstants(lookupConstraints).create()
```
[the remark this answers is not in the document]
jesuino: >  I would expect the "resize" settings to go away (so there's no confusion about why they don't work).

I agree, I created a ticket to address it https://issues.jboss.org/browse/AF-1790

However please notice that C3 will be the default and Google is marked to be deprecated soon, so in Business Central this should be a problem.

dashbuilder/dashbuilder-client/dashbuilder-renderers/dashbuilder-renderer-c3/src/main/java/org/dashbuilder/renderer/c3/mutationobserver/MutationObserverInitWrapper.java
jesuino: Created this: 
https://issues.jboss.org/browse/AF-1791

Created the following PRs: 

https://github.com/kiegroup/appformer/pull/606
https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/913
https://github.com/errai/errai/pull/372

Could we merge this PR and then after AF-1791 is resolved come back and create a new ticket?

Thanks

jhrcek: +1 to proceed this step by step (first merge this PR, then update elemental, then remove the workarounds needed for older elemental).

jesuino:
jenkins retest this

jesuino:
@ederign  is this good do merge *_* Could you please review it again?

Thanks!

PR #618, opened 2019-01-28
[AF-1795]: Improved cache naming

adrielparedes:
In ISPN server might exists other caches that are not created by AppFormer Backend, so now, those caches contains a prefix to identify what caches are from AppFormer and those who are not. This way search and cache creation will only be on platform created caches. Nothing changed on Indexing API.

### DIFF
--- a/uberfire-extensions/uberfire-metadata/uberfire-metadata-backends/uberfire-metadata-backend-infinispan/src/main/java/org/uberfire/ext/metadata/backend/infinispan/provider/InfinispanContext.java
+++ b/uberfire-extensions/uberfire-metadata/uberfire-metadata-backends/uberfire-metadata-backend-infinispan/src/main/java/org/uberfire/ext/metadata/backend/infinispan/provider/InfinispanContext.java
@@ -31,6 +31,7 @@ import org.infinispan.client.hotrod.RemoteCache;
 import org.infinispan.client.hotrod.RemoteCacheManager;
 import org.infinispan.client.hotrod.configuration.AuthenticationConfigurationBuilder;
 import org.infinispan.client.hotrod.configuration.ConfigurationBuilder;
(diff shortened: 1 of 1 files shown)


uberfire-extensions/uberfire-metadata/uberfire-metadata-backends/uberfire-metadata-backend-infinispan/src/main/java/org/uberfire/ext/metadata/backend/infinispan/provider/InfinispanContext.java
```diff
@@ -278,17 +297,22 @@ public void dispose() {
     }
 
     public List<String> getIndices() {
-        return new ArrayList<>(this.cacheManager.getCacheNames());
+        return new ArrayList<>(this.cacheManager.getCacheNames())
+                .stream()
+                .filter(s -> s.startsWith(CACHE_PREFIX))
+                .map(s -> s.replace(CACHE_PREFIX,
```
jhrcek: `s.replace(regularExression, replacement)` internally [compiles regular expression](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#replaceAll(java.lang.String,%20java.lang.String)) every time you call it.
This could be made significantly faster by just using `.map(s -> s.substring(CACHE_PREFIX.length()))`

adrielparedes:
Jenkins retest this

jhrcek:
@adrielparedes the flood of warnings was not fixed :) I just noticed new kind of warnings I didn't notice before. What I reported in RHPAM-1848 looks like something that could be easily fixed.

ederign:
Please wait for the green build after https://github.com/kiegroup/appformer/pull/620

adrielparedes:
Jenkins retest this

PR #619, opened 2019-01-29
[AF-1800]: Mockito version downgraded

adrielparedes:
ip-bom contains 2.x mockito version that is not compatible with our tests. A new Jira was created to do the alignment, meanwhile a downgrade is required to maintain compatibility.

@ederign @mbiarnes would you mind to review?


mbiarnes:
this PR was replaced by https://github.com/kiegroup/appformer/pull/620

PR #624, opened 2019-02-01
AF-1706: Dataset remains broken after incorrect filter values have been fixed

jesuino:


jesuino:
would you please review, @ederign @jhrcek 

Thanks.

dashbuilder/dashbuilder-client/dashbuilder-displayer-client/src/main/java/org/dashbuilder/displayer/client/AbstractDisplayer.java
```diff
@@ -319,6 +319,7 @@ public void notFound() {
                     @Override
                     public boolean onError(final ClientRuntimeError error) {
                         showError(error);
+                        drawn = false;
```
ederign: this code is also not so clear. Could you please give it some love?

jesuino: @ederign I added a test and hopefully made the change a little more clear now. Basically when an error is thrown the elements containing the displayed are removed, and it will only show the error even if we fix the issue. To solve this it is required to make the displayer draw itself again, and the flag drawn controls when draw() is called or not.

PR #628, opened 2019-02-06
[AF-1797]: Added branch to index name

adrielparedes:
The problem occur because branch name was not being concatenated to index name during initial set up. Also log message was improved because not real indexing happens at that point

@ederign @porcelli would you mind to review?


adrielparedes:
jenkins retest this

PR #629, opened 2019-02-07
[do not merge] RHPAM-1859:Horizontal scrollbars missing in task inbox

nmirasch:


nmirasch:
@cristianonicolai just a POC of a possible solution for the missing scrolling at some list with a lot of columns. The point is at least ensure 120px for any column allowing the user expand the width. In this commit I set that expandwidth button displayed by default, just to let you see how it can look like.
Please, let me know what you think and with your comment we can refine the implementation. Thanks!

PR #633, opened 2019-02-13
[AF-1603]: Decrease amount of ISPN Warn messages

adrielparedes:
Change the order protobuf fields are read.

@jhrcek @ederign would you mind to review?

### DIFF
--- a/uberfire-extensions/uberfire-metadata/uberfire-metadata-backends/uberfire-metadata-backend-infinispan/src/main/java/org/uberfire/ext/metadata/backend/infinispan/proto/KObjectMarshaller.java
+++ b/uberfire-extensions/uberfire-metadata/uberfire-metadata-backends/uberfire-metadata-backend-infinispan/src/main/java/org/uberfire/ext/metadata/backend/infinispan/proto/KObjectMarshaller.java
@@ -78,7 +78,6 @@ public class KObjectMarshaller implements MessageMarshaller<KObject> {
 
         List<KProperty<?>> properties = descriptor.getFields()
                 .stream()
-                .filter(fieldDescriptor -> !isMainAttribute(fieldDescriptor))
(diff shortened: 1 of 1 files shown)


adrielparedes:
@ederign those tests are in uberfire-metadata-io, they are still green. The only thing I did in this class is to change how the field was being retrieved. But all tests about behavior are in metadata-io because it needs protobuf to be tested.

uberfire-extensions/uberfire-metadata/uberfire-metadata-backends/uberfire-metadata-backend-infinispan/src/main/java/org/uberfire/ext/metadata/backend/infinispan/proto/KObjectMarshaller.java
```diff
@@ -105,6 +110,13 @@ public KObjectImpl readFrom(ProtoStreamReader protoStreamReader) throws IOExcept
                                !StringUtils.isEmpty(fullText));
     }
 
+    private String getAndRemove(final List<KProperty<?>> properties,
+                                final String key) {
+        KProperty<?> value = properties.stream().filter(kProperty -> kProperty.getName().equals(key)).findFirst().get();
```
jhrcek: @adrielparedes as discussed you said that all the properties are mandatory, so it is safe to use `Optinal.get()`.
But it might still be better to use `optional.orElseThrow(() -> new IllegalStateException("Missing mandatory property " + key));` to get better error message in case it fails for some unexpected reason and to avoid IntelliJ's warning "get without isPresent check"

adrielparedes:
@jhrcek I think 2 is related to an issue you found and I think it will be fixed together with RHPAM-1821

PR #635, opened 2019-02-14
RHPAM-1825: Reload displayer properties when renderer is changed

jesuino:


jesuino:
@ederign @jhrcek Would you please review? Now when a renderer is changed the settings is also reloaded. I only reload the settings when the renderer is changed, it feels weird update it all the time that every property is changed.

Thanks!

dashbuilder/dashbuilder-client/dashbuilder-displayer-client/src/main/java/org/dashbuilder/displayer/client/widgets/DisplayerEditor.java
```diff
@@ -367,9 +373,14 @@ void onDataSetLookupChanged(@Observes DataSetLookupChangedEvent event) {
     }
 
     void onDisplayerSettingsChanged(@Observes DisplayerSettingsChangedEvent event) {
+        String newRenderer = event.getDisplayerSettings().getRenderer();
```
jhrcek:
```suggestion
        String newRenderer = displayerSettings.getRenderer();
```

jesuino:
@jhrcek For some reason I am not seeing your comment on github, but I just made an improvement of checking possible current renderer before getting the default one. There are the cases where settings.getRenderer is empty hence why I was getting the default one, after your comment I noticed that user may have set a renderer before, hence why I added the additional checking.

Thanks.

PR #639, opened 2019-02-18
Fix ConcurrentModificationException in TaskListViewImplTest test

cristianonicolai:


cristianonicolai:
@nmirasch @jstastny-cz @ederign @mbiarnes  fixing tests on jbpm-wb that are currently failing on master

cristianonicolai:

nmirasch:
@jstastny-cz regarding master is broken and this PR fix the issue, maybe is a good idea merge this one, and probably add that test in a second step, what do you think?

porcelli:
@nmirasch disagree. tests are the best option to prevent future problems.

jstastny-cz:
@nmirasch @cristianonicolai I am fine with deferred addition of a test case (I hope we can agree that this should be really an exception rather than a rule)

uberfire-extensions/uberfire-widgets/uberfire-widgets-table/src/main/java/org/uberfire/ext/widgets/table/client/UberfireColumnPicker.java
```diff
@@ -145,12 +145,7 @@ public void addColumn(ColumnMeta<T> columnMeta) {
     }
 
     protected void sortAndAddColumns(List<ColumnMeta<T>> columnMetas) {
-        // Sort based on preferences applied
-        Collections.sort(columnMetas);
-        //Add the columns based on the preferences
-        for (ColumnMeta meta : columnMetas) {
-            addColumn(meta);
-        }
+        columnMetas.stream().sorted().forEach(meta -> addColumn(meta));
```
ederign: @cristianonicolai how this fix an ConcurrentModificationException?

nmirasch:
@porcelli absolutely  agree with you about the test as the best option, I just was thinking in the broken test that needs this PR code to get the green light. But it seems we have to wait for @cristianonicolai  action anyway.

porcelli:
if this is my call, I'd prefer to wait for the tests. as @ederign already asked, I can't see why the proposed change could fix the test... so I prefer to wait.

mbiarnes:
@porcelli @etirelli @krisv once you approve this please merge it when it is green - this is the last commit that is to cherry-pick for7.3 CR1

cristianonicolai:
@porcelli @ederign  test added

uberfire-extensions/uberfire-widgets/uberfire-widgets-table/src/main/java/org/uberfire/ext/widgets/table/client/UberfireColumnPicker.java
cristianonicolai: @ederign the same collection is modified inside the scope of _addColumn_, that's why using streams completely avoids that due to its immutable nature.

barboras7:

porcelli:
failing tests are the `well know` random ones recorded by https://janhrcek.cz/random-failures/

PR #642, opened 2019-02-20
[AF-1842]: Added 5 retries before showing disconnection error

adrielparedes:
When connection is reestablished, the retries counter is reset. In tests you can see the behavior

@porcelli Would you mind to review?


adrielparedes:
jenkins retest this

PR #653, opened 2019-02-27
[BAQE-888] Investigate a fix JGit community tests failures on windows

domhanak:
* Unlock locked 'doraFs' instance after the asserts
Locks make this test fail on Windows, because of how
Windows handles files in use. 
Also multiple tests can't be executed because `AbstractTestInfra` is unable to deleted locked files on Windows.

* Set file permisions differently when running on Windows
PosixFilePermission is not supported in Windows environment, adds a branch  that sets permissions in compatible way 

* Wrap path assertions strings with File to make it OS agnostic 

Use File to make asserts that check paths work on windows properly too


domhanak:
@ederign @pefernan Please check it out. In a meantime I am submitting PR to kie-docs to enhance post-commit feature documentation. After some investigation I found out that JGit does not support git-hook on windows without cygwin installed and properly configured. Basically you need to have cygwin installed with git package and your PATH variable must contain path to it  - `C:\cygwin64\bin\`

domhanak:
@jhrcek if you have time to review :)

domhanak:
https://github.com/kiegroup/kie-docs/pull/1418 mentioned doc PR

PR #667, opened 2019-03-18
DROOLS-3103 : Support run test at project level

Rikkola:
https://issues.jboss.org/browse/DROOLS-3103

@kkufova @danielezonca @gitgabrio If you could please take a look.

Bundle:
https://github.com/kiegroup/appformer/pull/667
https://github.com/kiegroup/kie-wb-common/pull/2568
https://github.com/kiegroup/drools-wb/pull/1115

### DIFF
--- a/uberfire-services/uberfire-services-backend/src/main/java/org/guvnor/common/services/backend/test/TestRunnerServiceImpl.java
+++ b/uberfire-services/uberfire-services-backend/src/main/java/org/guvnor/common/services/backend/test/TestRunnerServiceImpl.java
@@ -11,12 +11,8 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
-*/
+ */
 package org.guvnor.common.services.backend.test;
-
-import java.lang.annotation.Annotation;
-import java.util.ArrayList;
-import java.util.List;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.event.Event;
@@ -24,7 +20,6 @@ import javax.enterprise.inject.Any;
 import javax.enterprise.inject.Instance;
 import javax.inject.Inject;
 
-import org.guvnor.common.services.shared.test.Failure;
 import org.guvnor.common.services.shared.test.TestResultMessage;
 import org.guvnor.common.services.shared.test.TestRunnerService;
 import org.guvnor.common.services.shared.test.TestService;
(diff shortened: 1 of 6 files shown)


Rikkola:
@kkufova Not actually sure how to tests this as a standalone, since the bug this fixes only happens with multiple asset types. Might be best to wait for all the PRs.

uberfire-services/uberfire-services-backend/src/main/java/org/guvnor/common/services/backend/test/TestRunnerServiceImpl.java
```diff
@@ -60,18 +60,43 @@ public void runAllTests(final String identifier,
     public void runAllTests(final String identifier,
                             final Path path,
                             final Event<TestResultMessage> customTestResultEvent) {
-
-        final TestRunEventCollection testResultEvent = new TestRunEventCollection();
+        final ArrayList<TestRunEventCollection> events = new ArrayList<>();
 
         for (final TestService testService : testServices) {
+            final TestRunEventCollection testResultEvent = new TestRunEventCollection();
```
gitgabrio: @Rikkola 
HI, I'm reading on browser, so maybe misreading. 
Anyway, it seems to me that inside loop at line 65 you populate a collection, then with methods called at lines 73-75 you iterate the same collection three times (one for method) to calculate the parameter used inside the fire event:  would not be possible to do such calculation directly inside the loop at line 65 ? Not sure it is worth anything, maybe just if the input is big

Rikkola: The collection size will probably never go above 5, but better to be safe just sum them up on the go. I'll do change.

Rikkola:
Just as note to myself. I'll hold the other PRs for this at the moment. There is an issue with resetting the empty status when the user has ran "run all" and then opens an editor that also uses the test run results panel.
Might be best to merge the docks fix in first.

uberfire-services/uberfire-services-backend/src/main/java/org/guvnor/common/services/backend/test/TestRunnerServiceImpl.java
Rikkola: Fixed.

Rikkola:
@gitgabrio Actually now that I did the change on how the tests are ran for single test file I'm pretty sure I can remove `TestRunEventCollection` taking a look.

gitgabrio:
@Rikkola 
I've re-read the code - it seems previously I had misunderstood/misread something. Please let me know when you remove the TestRunEventCollection

uberfire-services/uberfire-services-backend/src/main/java/org/guvnor/common/services/backend/test/TestRunEventCollection.java
```diff
@@ -0,0 +1,69 @@
...
+import java.lang.annotation.Annotation;
+import java.util.ArrayList;
+import java.util.List;
+
+import javax.enterprise.event.Event;
+
+import org.guvnor.common.services.shared.test.Failure;
+import org.guvnor.common.services.shared.test.TestResultMessage;
+
+public class TestRunEventCollection
```
danielezonca: Can you please rename it to `TestRunResultCollector` or `TestRunResultAggregator`? I think *Collection is really misleading as name

uberfire-services/uberfire-services-backend/src/main/java/org/guvnor/common/services/backend/test/TestRunEventCollection.java
```diff
@@ -0,0 +1,69 @@
...
+        implements Event<TestResultMessage> {
```
danielezonca: To be honest I don't like to implement an J2EE annotation as "fake" class to aggregate the result. I understood you did this to reuse the same `runTest` method but I think it is more understandable to create a new wrapper/interface like `TestResultConsumer` that accept `TestResultMessage`. Then create two implementation of that consumer, one `AccumulatorConsumer` and one `CdiEventConsumer` that do the proper firing.
Wdyt? It is also fine to keep it as is but add at least some javadoc to this class

Rikkola: Yeah I moved away from this with the recent update. The events are no longer used to report a single test run for example, so no need to abuse the Event.

Rikkola:
@gitgabrio ok I moved away from the Test abuse and now just returning lists instead of collecting the result with the event.

PR #678, opened 2019-04-04
DROOLS-3438: Unify cell editing cancellation 

jomarko:
Ensemble together with:
- kiegroup/drools-wb#1112

This PR unifies the way how Guided Decision Table and Scenario Simulation user can cancel the cell edit mode.

- By enter, new value is stored, selection is moved down
- By shift + enter, new value is stored, selection is moved up
- By tab, new value is stored, selection is moved right
- By shift + tab, new value is stored, selection is moved left
- By Esc, new value is discarded, selection remains same

### DIFF
--- a/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dom/single/impl/BaseSingletonDOMElementFactory.java
+++ b/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dom/single/impl/BaseSingletonDOMElementFactory.java
@@ -74,7 +74,11 @@ public abstract class BaseSingletonDOMElementFactory<T, W extends Widget, E exte
 
     @Override
     public void destroyResources() {
-        flush();
+        if (e != null) {
+            e.detach();
+            widget = null;
+            e = null;
+        }
     }
 
     @Override
--- a/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/keyboard/KeyDownHandlerCommon.java
+++ b/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/keyboard/KeyDownHandlerCommon.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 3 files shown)


uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dom/single/impl/BaseSingletonDOMElementFactory.java
```diff
@@ -74,7 +74,11 @@ public void execute() {
 
     @Override
     public void destroyResources() {
-        flush();
+        if (e != null) {
```
manstis: `destroyResources()` will also be called when, for example, resizing a column. 

This will now cancel the change. Is this the expected behaviour?

jomarko: Well, I am am not aware of specification how such case should behave, should I ask UX team? The problem of these `destroyResources` and `flush` I touched is in they repetition, will start a conversation ,on email probably, where will try to summarize my findings.

manstis: OK. It is probably something for UX to determine.

I just mentioned it really to highlight `destroyResources()` is not used solely by keyboard operations.

jomarko: @manstis yes, however still think we should have this behavior implemented:
- destroyResources: don't care about value in widget
- flush: store value in widget

I mean, if we want to keep current behavior, I rather vote fore replacing current `destroyResources` usages by `flush`.

manstis: @jomarko I don't have any issues with this PR.. I was just highlighting. I tend to agree, that use of `<Enter>` or `<Tab>` to _flush_ the value is correct and that resizing a column should _discard_ the change.. green light on this PR from me.

jomarko: @manstis seems Liz have different opinion. She said values are expected to remain if column resized. I need to update the PR.

jomarko:
@manstis I think we can merge

PR #685, opened 2019-04-23
DROOLS-3591: [DMN Designer] Data Types - Type dropdown is not navigable by keyboard

manstis:
See https://issues.jboss.org/browse/DROOLS-3591

Also see https://issues.jboss.org/browse/RHDM-912

The problem was that the `BootstrapSelect` Java Script was being loaded multiple times and event handlers for the component were being registered an equal number of multiple times. This lead to, for example, a cursor-down key press moving the selection multiple times.

I put a break point in `bootstrap-select.js` (L#1506) and observed the selection move one position, then another and another; for each successive registered event handler. Debugging `PatternFlyBootstrapper` showed the check failed and multiple instances of the `BootstrapSelect` were injected into the header. I also examined the `global` Java Script objects on the page and there was none for `Selectpicker` (nor `DateRangePicker` -- another script). 

This leads me to suspect GWT is obfuscating the resources in `PatternFlyClientBundle` and hence the existing checks unsuitable. That said, if anybody can think of a better solution than that posed in this PR I'm happy to try, test and modify :-) 

@jomarko This will probably need a test of Business Central.. I checked DMN and it seems to work fine to me.. however IDK where else the other libraries are used and hence it might be worth asking around (in QE) for others to check the BC WAR too to verify nothings broken. I'll leave that decision to you.

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/sys/PatternFlyBootstrapper.java
+++ b/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/sys/PatternFlyBootstrapper.java
@@ -26,125 +26,89 @@ import static org.uberfire.client.views.pfly.sys.MomentUtils.setMomentLocale;
  */
 public class PatternFlyBootstrapper {
 
+    private static boolean isjQueryLoaded = false;
+
+    private static boolean isPrettifyLoaded = false;
+
+    private static boolean isBootstrapSelectLoaded = false;
+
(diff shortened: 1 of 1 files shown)


manstis:
@jomarko I've reinstated the `jQuery` loading mechanism (as `jQuery` **was** on `$wnd` when I debugged). This leaves only the `PatternFlyClientBundle` resources using static flags. I really don't understand why `GwtBootstrap3ClientBundle` resources work but `PatternFlyClientBundle` do not.. IDK if @cristianonicolai or @ederign have any insight.

ederign:
hi @manstis I've no idea. :-)

PR #690, opened 2019-05-03
DROOLS-3940: DMN Editor: Update contextual menus (headers and cells)

manstis:
See https://issues.jboss.org/browse/DROOLS-3940

(Well, not directly related, but I found whilst writing DROOLS-3940 that deleting a column that had a selected header cell lead to strange behaviour - as the _selected header cells_ were not being cleared when the column deleted).


manstis:
@jomarko I've added support for keyboard activation of Context Menus in the Header.

You might want to re-review.

PR #696, opened 2019-05-10
[DROOLS-3993] Add "val" method to JQuerySelectPicker

gitgabrio:
@manstis @jomarko @karreiro 

This is the twin part of https://github.com/kiegroup/kie-wb-common/pull/2632
It add "val" method that is used in the other PR


gitgabrio:
@manstis @jomarko @karreiro
Please note: I've no idea how to write a *real* test for this - I would if it would possible to instantiate a real gwt environment, but not with mock

gitgabrio:
@manstis
Hi! Could you please merge this, squashing commits? Thanks!

PR #700, opened 2019-05-16
AF-1966: Changing from ManagedBean back to newInstance

jesuino:
ManagedBean respects the ApplicationScoped annotation, which means it always returns the same instance. In LayoutEditorPropertiesPresenter instances of LayoutElementPropertiesPresenter - if instances of LayoutElementPropertiesPresenter are created with ManagedInstance, then only a single one will be created. With newInstances the cache is correctly created for each element.


manstis:
It's not for me to say, but perhaps `LayoutElementPropertiesPresenter` should not be `@ApplicationScoped` in this instance!?! Just a comment when I saw this PR submitted.

jesuino:
That's a good point, @manstis  - AFAIK it is only used in LayoutEditorPropertiesPresenter, could it use Dependent scope?

manstis:
@jesuino Given the limited knowledge I have of this area and what you've said; I'd say "yes".

paulovmr:
@jesuino I agree with @manstis , I think it is worth testing changing it to `@Dependent`, probably will be enough to fix the problem.

jesuino:
thanks @paulovmr @manstis  your suggestion worked!

PR #702, opened 2019-05-20
upgraded src/main/config/zanata.xml to link to new project 2.20.0 on https://vendors.zanata.redhat.com/

kiereleaseuser:


mbiarnes:
the script for generating new Zanata versions has to be executed again - this PR will be closed as it will be raised again

PR #723, opened 2019-07-01
[AF-2053]: Notification is thrown once the project is created in FS.

adrielparedes:
# Problem
When a project was deleted, the notification was thrown before FS deletion, so the refresh was execute before any information was deleted and the UI was not refreshed because of that.

# Solution
The solution is quite simple, the Notification is thrown after the FS is created. That way, we ensure that information is present before the refresh event arrives to other nodes in the cluster. This also fix delete modal problem.

@pefernan  @paulovmr  @ederign would you mind to review?


adrielparedes:
jenkins retest this

adrielparedes:
@paulovmr I've added that test, please let me know if that way is ok. Thanks for the comment

tomasdavidorg:
Please do not merge. We need to first check the fixes on the latest OpenShift images.

PR #731, opened 2019-07-05
AF-2053: Dialog remains open after asset is deleted at the same time

pefernan:
The issue was caused when two users were trying to delete the same asset at same time (with the delete popup opened). When first user finally deletes the asset the second user gets the concurrent delete notification, whatever option he chooses (cancel or ignore) on that popup, the delete popup will still be opened, so he will be able click on the delete button, getting that error.

To prevent that situation, we'll be hiding the delete popup if there's a concurrent deletion.

@tomasdavidorg @adrielparedes @ederign Could you please take a look?

This is part of an ensemble, please merge with:
https://github.com/kiegroup/appformer/pull/731
https://github.com/kiegroup/kie-wb-common/pull/2740


pefernan:

pefernan:
jenkins execute full downstream build please

pefernan:
jenkins execute full downstream build please

pefernan:
jenkins execute full downstream build please

PR #739, opened 2019-07-12
RHPAM-2036 :Group cannot be added in Business central integrated with…

RishiRajAnand:
… RHSSO integration

- We are getting this error "java.lang.NoSuchMethodError:org.keycloak.representations.idm.RoleRepresentation.setScopeParamRequired(Ljava/lang/Boolean;)V" because at runtime appformer is calling method setScopeParamRequired in keycloak-core with a primitive boolean, while the method itself accept a Boolean wrapper type.
- Added Boolean wrapper type as method argument.
add-group-rhsso

### DIFF
--- a/uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-security-management-keycloak/src/main/java/org/uberfire/ext/security/management/keycloak/KeyCloakGroupManager.java
+++ b/uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-security-management-keycloak/src/main/java/org/uberfire/ext/security/management/keycloak/KeyCloakGroupManager.java
@@ -119,7 +119,7 @@ public class KeyCloakGroupManager extends BaseKeyCloakManager implements GroupMa
             final RoleRepresentation roleRepresentation = new RoleRepresentation();
             roleRepresentation.setName(entity.getName());
             roleRepresentation.setDescription(entity.getName());
-            roleRepresentation.setScopeParamRequired(false);
+            roleRepresentation.setScopeParamRequired(new Boolean(false));
             roleRepresentation.setId(entity.getName());
             roleRepresentation.setComposite(false);
             final ClientResponse response = (ClientResponse) rolesResource.create(roleRepresentation);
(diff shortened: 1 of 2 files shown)


domhanak:
@RishiRajAnand Could  you please also add test for create method? :) I do not see it covered in `KeyCloakGroupManagerTest`

RishiRajAnand:
@domhanak Sure. I'll add that too

RishiRajAnand:
> @RishiRajAnand Could you please also add test for ` public Group create(Group entity)` method? :) I do not see it covered in `KeyCloakGroupManagerTest`

@domhanak Pushed the test case for creating group.

domhanak:

domhanak:
jenkins execute compile downstream build

domhanak:
jenkins retest this

domhanak:
Compile downstream build is failing because machine has no space left.

domhanak:
jenkins retest this

domhanak:
I have used JAR with the fix and still could reproduce the behavior. Could some1 else give it a try?

krisv:
I would at least expect a commit that would upgrade the version of keycloak used, which would result in this code no longer being able to compile as the operation has been removed (see description I left in JIRA).  @ederign is there someone who might be able to help Rishiraj?

RishiRajAnand:
@krisv Sorry I missed your comment in JIRA. Yes the newer version of key cloak (4.8.3) has removed this method  **setScopeParamRequired** and due to that it gives an error when used  with RH-SSO 7.3 which uses keycloak version 4.8.3. I was testing with version RH-SSO 7.2 due to which i was not able to replicate it the same way in my system. Given that, The best way in this scenario could be to upgrade keycloak to version 4.8.3. But that would also require the existing clients to upgrade to RH-SSO 7.3. wdyt?

mareknovotny:
Guys, I see this is targeting 7.23.x branch. Is there a commit already on master for this? If you plan to upgrade keycloak or any dependency this should go into master and not to 7.23.x where the upgrades are generally not possible as this is stable supported branch.
I would recommend to do first master branch change and do the backporting too.

RishiRajAnand:
Closing this pull request since the the new keycloak upgrade to 4.8.3 is targeting for master branch. Sending another PR with new changes.Thanks @mareknovotny @domhanak @krisv for your inputs on this one :)

PR #766, opened 2019-08-12
AF-2138: initialized the repositories on app initialization

porcelli:
making the repos available via http, ssh or git without requiring any UI interaction


domhanak:
Jenkins execute full downstream build!

PR #774, opened 2019-08-19
[AF-2049]: Added retries to fix slow connection problems

adrielparedes:
@ederign would you mind to review?


uberfire-extensions/uberfire-metadata/uberfire-metadata-backends/uberfire-metadata-backend-infinispan/src/main/java/org/uberfire/ext/metadata/backend/infinispan/provider/InfinispanContext.java
```diff
@@ -311,24 +312,27 @@ public void addProtobufSchema(String clusterId,
...
-                        marshaller.registerSchema(protoTypeName,
-                                                  proto,
-                                                  KObject.class);
-                    } catch (IOException e) {
-                        throw new InfinispanException("Can't add protobuf schema <" + protoTypeName + "> to cache",
-                                                      e);
-                    }
-                });
+
+        new Retry(5, () -> {
```
jakubschwan: What will happened if we run out of retries? IMO there should be some logging or/and exception that loading protobuf schema failed.

PR #785, opened 2019-08-30
[BAPL-1015] Workspaces Collaboration: Change Request

caponetto:
**Summary**
As a user, I want to be able to submit a change request from a branch of Business Central to another branch in the same project and also be able to review/approve changes from others to my project. This change request will contain all files from my project/branch.

**Goals**
Only textual file diffs will be supported.

**Non goals**
Conflict resolution is out of the scope of this requirement.

Further details available at [BAPL-1015](https://issues.jboss.org/browse/BAPL-1015) and its tasks.

**Associated pull requests**
https://github.com/kiegroup/jbpm-wb/pull/1383
https://github.com/kiegroup/optaplanner-wb/pull/346
https://github.com/kiegroup/drools-wb/pull/1224
https://github.com/kiegroup/kie-wb-distributions/pull/963
https://github.com/kiegroup/kie-wb-common/pull/2850
https://github.com/kiegroup/appformer/pull/785

@ederign 
@paulovmr


caponetto:
I'm checking SonarCloud issues.

domhanak:

caponetto:

domhanak:
@caponetto I checked the SonarCloud Code Analysis and the remaining coverage failures can probably be ignored. Most of them are POJO's so I think the coverage looks good. Thank you for using it to our benefit.

ederign:
@barboras7 could you please take a look at this

barboras7:
@ederign Sure, I'm working on it. However, it's quite a big PR so it is going to take a while. I started manual testing today and will publish some info as soon as I have it prepared.

barboras7:
Hi @caponetto, @ederign, 

I did some manual checks today and found some issues. Some of them are just suggestions/questions, see the list below:

1. When you create a branch from master, on this branch you change the version in GAV in Settings (Settings tab -> Version) and then create a change request with this change, you can see multiple changed files there. One of them is package-white-list but there is no obvious change in that file.
2. User with contributor role is able to accept/reject change request. In my opinion, only some users with higher permissions should be able to do it (at least, that's what I thought would be the biggest benefit of this feature)
3. When a change request is rejected, it can not be accepted anymore. Therefore, I think there might be some alert like "If you reject this change request..." because users might think, that rejecting is something similar to "request changes".
4.  When there is one change request and then it is accepted/rejected, the screen with open change requests says "No results found". It should probably say the same thing as in the beginning "Nothing here. There are currently no change requests..."
5. This is just a small observation but I think that for most parts in the business central, we use the convention "Add xy" (Add project, Add asset, Add exception) so maybe instead of "Comment", we could use "Add comment"
6. I did not find any option to delete a change request. Of course, it can be rejected/accepted and then becomes "closed", but no way to delete it (if a user opens a change request and decide they want to close)

I will continue with the review tomorrow. Let me know if you have any questions, ideas, or what you think about what I wrote

caponetto:
Hi @barboras7, thanks for this! Follow my answers.

1. I've noticed that too. However, we just show the output of a _git diff_ command. If you execute a _git diff_ in the terminal having this scenario, you would see the same differences, including for package-white-list. Perhaps this file is created in a way and updated differently, thus causing the diff. Anyway, I will check if we can do something about it.

2. Our rule for this case is the following: if the user has _write_ permission on the repository or destination branch, then they can accept/reject the change request. This is just to simplify things for this first version, but we can definitely refine that in the future.

3. How about we include a confirmation popup for this case?

4. For that case, I've followed the same behavior found in the Assets tab. "Nothing here" is only shown when there is absolutely nothing there - you can't even see the filters/pagination. On the other hand, if you have a bpmn and filter the list to "Optimization", then you would see "No results found". That's similar with change requests, because, for the scenario you've described, there are no Open change requests and the filter is set to "Open".

5. Sure, I'll include this change in the next commits.

6. We can provide a close button only available for the author while the change request is Open.
Then, if such action is performed, the status would become something like "closed by the author". What do you think?
A real delete action, meaning a complete removal from the lists, is only done if the user deletes one of the involved branches (source or target).

@paulovmr let us know if you think differently.

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/repositories/ChangeRequestServiceImpl.java
```diff
@@ -0,0 +1,1256 @@
...
+import static java.lang.Integer.min;
+import static java.util.stream.Collectors.toMap;
+import static org.kie.soup.commons.validation.PortablePreconditions.checkNotEmpty;
+import static org.kie.soup.commons.validation.PortablePreconditions.checkNotNull;
+
+@Service
+@ApplicationScoped
+public class ChangeRequestServiceImpl implements ChangeRequestService {
+
+    // Avoiding string duplication - SonarCloud
```
barboras7: I think this comment is not necessary. Please, delete it of that's the case

uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/util/commands/BranchUtil.java
```diff
@@ -43,4 +45,12 @@ public static void deleteUnfilteredBranches(final Repository repository,
                 .setForce(true)
                 .call();
     }
+
+    public static void existsBranch(final Git git,
+                                    final String branch) {
+        if (git.getRef(branch) == null) {
+            throw new GitException(String.format("Branch <<%s>> does not exists",
```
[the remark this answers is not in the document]
caponetto: Thanks! I'll fix it.

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/repositories/ChangeRequestServiceImpl.java
caponetto: I'll delete this.

uberfire-nio2-backport/uberfire-nio2-impls/uberfire-nio2-jgit/src/main/java/org/uberfire/java/nio/fs/jgit/util/commands/GetCommonAncestorCommit.java
```diff
@@ -0,0 +1,58 @@
...
+
+import org.eclipse.jgit.revwalk.RevCommit;
+import org.eclipse.jgit.revwalk.RevWalk;
+import org.eclipse.jgit.revwalk.filter.RevFilter;
+import org.uberfire.java.nio.fs.jgit.util.Git;
+import org.uberfire.java.nio.fs.jgit.util.exceptions.GitException;
+
+import static org.kie.soup.commons.validation.PortablePreconditions.checkNotNull;
+
+public class GetCommonAncestorCommit {
```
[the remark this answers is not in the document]
caponetto: They are commands, so the name must imply some action. We already have, for example, GetFirstCommit, GetLastCommit, GetRef, among others.

caponetto:

caponetto:

caponetto:
Hi @barboras7, I've updated this PR, can you take a look when you have some time? Thanks!

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/repositories/RepositoryServiceImpl.java
```diff
@@ -450,6 +450,8 @@ protected void doRemoveRepository(final OrganizationalUnit orgUnit,
             //Remove reference to Repository from Organizational Units
             for (Repository repository : orgUnit.getRepositories()) {
                 if (repository.getAlias().equals(alias)) {
+                    configStorage.deleteRepository(alias);
```
paulovmr: This method should not be called here (and I think it shouldn't be needed anywhere else). Here, the repository is only marked for deletion by the `organizationalUnitService` (with lock acquired), and later the `FileSystemDeleteWorker#removeRepository(final Repository repo)` will remove the whole repository directory at once, which should remove the change requests along with it.

caponetto: Thanks for pointing this out! I guess  `FileSystemDeleteWorker` wasn't there yet when I changed this method the first time. Let me check if we still need it.

caponetto:

caponetto:

caponetto:
jenkins retest this

barboras7:
Thanks @caponetto, great work!

caponetto:

caponetto:

PR #786, opened 2019-08-30
[AF-2182]: Added generation of base caches in case they don't exist

adrielparedes:
When ISPN got rebooted the main caches were lost, also the indexes because ISPN is not a persistent cache. So when ISPN is restarted, BC generates caches again and force reindexing.


adrielparedes:
jenkins retest this

PR #794, opened 2019-09-12
BXMSPROD-525 remove version.org.kie property

Ginxo:
https://issues.jboss.org/browse/BXMSPROD-525

the version `version.org.kie` property is duplicated what we have in kie-parent, so it is now useless https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/pom.xml#L343


mareknovotny:
Jenkins execute full compile downstream build

PR #797, opened 2019-09-13
[AF-2232] IllegalStateException when logging out from BC

gabriel376:
**[AF-2232](https://issues.jboss.org/browse/AF-2232) IllegalStateException when logging out from BC**

**Issue:**
* Some activities are not `closed` before calling `destroy` in shutdown (logout) operation

**Solution:**
* Always call `close` before `destroy`

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client/src/test/java/org/uberfire/client/workbench/panels/impl/PlaceManagerTest.java
+++ b/uberfire-workbench/uberfire-workbench-client/src/test/java/org/uberfire/client/workbench/panels/impl/PlaceManagerTest.java
@@ -423,6 +423,16 @@ public class PlaceManagerTest {
     }
 
     @Test
+    public void testClosePlaceAlwaysCloseActivityBeforeDestroy() {
+        when(kansasActivity.isType(any())).thenReturn(false);
+
+        placeManager.closePlace(kansas);
+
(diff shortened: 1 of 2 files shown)


uberfire-workbench/uberfire-workbench-client/src/test/java/org/uberfire/client/workbench/panels/impl/PlaceManagerTest.java
```diff
@@ -422,6 +422,16 @@ public void testNormalCloseExistingScreenActivity() throws Exception {
...
     }
 
+    @Test
+    public void testClosePlaceAlwaysCloseActivityBeforeDestroy() {
+        when(kansasActivity.isType(any())).thenReturn(false);
+
+        placeManager.closePlace(kansas);
+
+        verify(kansasActivity).onClose();
+        verify(activityManager).destroyActivity(kansasActivity);
```
domhanak: Would it be possible to verify the order of calls too? Like that `onClose()` was really executed before `destroy` was called? Not a hard req, just thinking that it could give better coverage.

gabriel376: good point! I sent a new commit to cover this, thank you!

PR #808, opened 2019-09-30
BXMSPROD-525 version.org.kie property removed

Ginxo:
https://issues.jboss.org/browse/BXMSPROD-525

the version version.org.kie property is duplicated what we have in kie-parent, so it is now useless https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/pom.xml#L343


Ginxo:
replaced by https://github.com/kiegroup/appformer/pull/833

PR #800, opened 2019-09-20
[7.5.1] [AF-2214] User confirmation when closing the workbench with unsaved stuff

caponetto:
Same code: https://github.com/kiegroup/appformer/pull/796

JIRA task: [AF-2214](https://issues.jboss.org/browse/AF-2214)

If the user has unsaved changes and closes the browser, then they lost all changes without being warned about. Thus, as a last resort, in this PR the browser will ask the user for confirmation when closing the tab, closing the browser or refreshing the page.

**DEMO**
AF-2214-2019-09-10

PR #796, opened 2019-09-12
[AF-2214] User confirmation when closing the workbench with unsaved stuff

caponetto:
JIRA task: [AF-2214](https://issues.jboss.org/browse/AF-2214)

If the user has unsaved changes and closes the browser, then they lost all changes without being warned about. Thus, as a last resort, in this PR the browser will ask the user for confirmation when closing the tab, closing the browser or refreshing the page.

**DEMO**
AF-2214-2019-09-10

PR #819, opened 2019-10-16
AF-2283: Cannot open standalone perspective in Firefox

paulovmr:


paulovmr:

manstis:
@paulovmr @ederign IDK for sure, but if the test started failed recently could it be down to https://github.com/kiegroup/appformer/pull/796 that now prompts the User before closing the window? Please feel free to ignore me.

paulovmr:
@manstis Based on the behaviour I saw, the problem was indeed caused by  https://github.com/kiegroup/appformer/pull/796 , but not because of the prompt. It was because the `workbenchCloseHandler` started to be called by the `Window.addCloseHandler` instead of the `Window.addWindowClosingHandler` (that now shows the prompt). Apparently Firefox calls the close handler when the url is changed manually in the browser, and not only when the window is closed, like Chrome does, which causes the problem.

caponetto:
Thanks for this, @paulovmr!
I'll include the changes in https://github.com/kiegroup/appformer/pull/800 as well.

PR #829, opened 2019-10-29
[AF-2244] UX support/guidance after new deployment

gabriel376:
Implement [AF-2244](https://issues.jboss.org/projects/AF/issues/AF-2244) UX support/guidance after new deployment

Depends on https://github.com/kiegroup/kie-wb-common/pull/2986

Peek 2019-10-29 18-48

### DIFF
--- a/uberfire-api/src/main/java/org/uberfire/workbench/events/NotificationEvent.java
+++ b/uberfire-api/src/main/java/org/uberfire/workbench/events/NotificationEvent.java
@@ -31,7 +31,7 @@ public class NotificationEvent implements UberFireEvent {
     private Integer initialTopOffset;
     private boolean autoHide = true;
     private String navigationText;
-    private String navigationPath;
+    private PlaceRequest navigationPlace;
 
     public NotificationEvent() {
     }
@@ -123,25 +123,23 @@ public class NotificationEvent implements UberFireEvent {
     }
 
     public NotificationEvent setNavigation(final String text,
-                                           final String path) {
-        navigationText = text;
-        navigationPath = path;
+                                       final PlaceRequest place) {
+        this.navigationText = text;
+        this.navigationPlace = place;
         return this;
     }
 
     public boolean hasNavigation() {
         return navigationText != null
-            && !navigationText.isEmpty()
-            && navigationPath != null
-            && !navigationPath.isEmpty();
+            && navigationPlace != null;
     }
 
     public String getNavigationText() {
         return navigationText;
     }
 
-    public String getNavigationPath() {
-        return navigationPath;
+    public PlaceRequest getNavigationPlace() {
+        return navigationPlace;
     }
 
     @Override
--- a/uberfire-api/src/main/java/org/uberfire/workbench/events/NotificationEvent.java
+++ b/uberfire-api/src/main/java/org/uberfire/workbench/events/NotificationEvent.java
@@ -29,6 +29,9 @@ public class NotificationEvent implements UberFireEvent {
     private boolean isSingleton;
     private PlaceRequest placeRequest;
     private Integer initialTopOffset;
+    private boolean autoHide = true;
+    private String navigationText;
+    private String navigationPath;
 
     public NotificationEvent() {
     }
@@ -82,6 +85,7 @@ public class NotificationEvent implements UberFireEvent {
                              final boolean isSingleton,
                              final PlaceRequest placeRequest,
                              final Integer initialTopOffset) {
+
         this.notification = notification;
         this.type = type;
         this.isSingleton = isSingleton;
@@ -107,6 +111,37 @@ public class NotificationEvent implements UberFireEvent {
 
     public Integer getInitialTopOffset() {
         return initialTopOffset;
+    }
+
+    public NotificationEvent setAutoHide(final boolean autoHide) {
+        this.autoHide = autoHide;
+        return this;
+    }
+
+    public boolean autoHide() {
+        return autoHide;
+    }
+
+    public NotificationEvent setNavigation(final String text,
+                                           final String path) {
+        navigationText = text;
+        navigationPath = path;
+        return this;
+    }
+
+    public boolean hasNavigation() {
+        return navigationText != null
+            && !navigationText.isEmpty()
+            && navigationPath != null
+            && !navigationPath.isEmpty();
+    }
+
+    public String getNavigationText() {
+        return navigationText;
+    }
+
+    public String getNavigationPath() {
+        return navigationPath;
     }
 
     @Override
--- a/uberfire-project/uberfire-project-client/pom.xml
+++ b/uberfire-project/uberfire-project-client/pom.xml
@@ -119,12 +119,6 @@
     </dependency>
 
     <dependency>
-      <groupId>org.kie.workbench.screens</groupId>
-      <artifactId>kie-wb-common-workbench-client</artifactId>
-      <version>${project.version}</version>
-    </dependency>
-
-    <dependency>
       <groupId>org.jboss.errai</groupId>
       <artifactId>errai-ioc</artifactId>
     </dependency>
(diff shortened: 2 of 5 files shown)


paulovmr:
Hi @lclay2 , could you confirm if the deploy notification should not disappear after 2 seconds as the others (as can be seen in the gif)?

uberfire-api/src/main/java/org/uberfire/workbench/events/NotificationEvent.java
```diff
@@ -29,6 +29,9 @@
     private boolean isSingleton;
     private PlaceRequest placeRequest;
     private Integer initialTopOffset;
+    private boolean autoHide = true;
```
paulovmr: I think it would be better to set these new attributes in a constructor instead of using `set`methods.

gabriel376: I don't think it's a good idea.
There are already 6 constructors, and if I add a new one I have to set all default parameters of previous constructors. Default values should be set by class initialization, not by external classes, and if default values change in future, we have to change in all objects creation too.
Also, a constructor with 8 parameters (current 5 + new 3) is not a good idea.

Example:
```Java
// if I want to set only A and E, I must set also B, C and D
// even though I don't know what is the default value of these parameters
public class X {
    X(String A, String B, String C, String D, String E) { }
}

// when parameters have different types we can create different constructors
// with same length of parameters
// This is not possible when parameters have same type
public class X {
    X(int A) { }
    X(String B) { }
    X(int A, String B) { } // this is fine
    X(int A, String C) { } // this is not possible, don't know which constructor to call
}

// Problems:
//     don't want to set all parameters
//     it's not my responsibility
//     if default values changes in class impl, we have to change here too
//     constructor is too big
X x = new X("A", ?, ?, ?, "E");

// this is much better or something else like Builder Pattern
X x = new X().setA("A").setB("B");
```

paulovmr: @gabriel376 Ok, I agree with you. Maybe when we have named parameters and default values in Java we can change that. :)

uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/notifications/NotificationPopupsManagerView.java
```diff
@@ -77,14 +84,23 @@ public NotificationPopupHandle show(NotificationEvent event,
         view.setNotificationWidth(getWidth() + "px");
 
         if (event.hasNavigation()) {
-            view.addNavigation(event.getNavigationText(), () -> {
-                hideCommand.execute();
-                placeManager.goTo(event.getNavigationPath());
-            });
+            ResourceRef resourceRef = new ResourceRef(
```
paulovmr: I think a better approach to keep this permission check more flexible is to receive a `Promise<Boolean>` in the `setNavigation` method, and check it here. Mainly because there could be a case where the navigation needs additional checks to be added or not, or isn't even a perspective.

lclay2:
@paulovmr - I didn't work on this, so I need to pass your question to @amysueg to follow up on.

ederign:
@lclay2 @amysueg @gabriel376 @paulovmr I'll put my questions on the JIRA.

uberfire-api/src/main/java/org/uberfire/workbench/events/NotificationEvent.java
```diff
@@ -109,6 +113,37 @@ public Integer getInitialTopOffset() {
...
+        this.autoHide = autoHide;
+        return this;
+    }
+
+    public boolean autoHide() {
+        return autoHide;
+    }
+
+    public NotificationEvent setNavigation(final String text,
+                                           final String path) {
```
paulovmr: To keep it more flexible, it is better to change this path to a PlaceRequest.

barboras7:

uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/notifications/NotificationPopupsManagerView.java
```diff
@@ -70,7 +86,27 @@ public NotificationPopupHandle show(NotificationEvent event,
         view.setNotification(event.getNotification());
         view.setType(event.getType());
         view.setNotificationWidth(getWidth() + "px");
-        view.show(hideCommand);
+
+        if (event.hasNavigation()) {
+            final SyncBeanManager beanManager = IOC.getBeanManager();
+            final String identifier = event.getNavigationPlace().getIdentifier();
+            final ActivityBeansCache activityBeansCache = beanManager.lookupBean(ActivityBeansCache.class).getInstance();
```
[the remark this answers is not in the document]
gabriel376: I took this example by looking at other classes
Thanks for letting me know

gabriel376:

paulovmr:

barboras7:

barboras7:
@gabriel376 Hi,
could you please check the failures of builds? All of them are failing.

uberfire-project/uberfire-project-client/pom.xml
```diff
@@ -118,6 +118,12 @@
       <artifactId>uberfire-services-api</artifactId>
     </dependency>
 
+    <dependency>
```
paulovmr: This dependency shouldn't be added, it's the reason of the build failure.

uberfire-project/uberfire-project-client/src/main/java/org/guvnor/common/services/project/client/security/ProjectController.java
```diff
@@ -41,7 +41,12 @@
 import org.jboss.errai.common.client.api.Caller;
 import org.jboss.errai.security.shared.api.identity.User;
 import org.uberfire.client.promise.Promises;
+import org.uberfire.rpc.SessionInfo;
+import org.uberfire.security.ResourceRef;
 import org.uberfire.security.authz.AuthorizationManager;
+import org.uberfire.workbench.model.ActivityResourceType;
+
+import static org.kie.workbench.common.workbench.client.PerspectiveIds.SERVER_MANAGEMENT;
```
paulovmr: I missed this in my previous review, you can't depend on kie-wb-common in appformer. In this case, this new method should be somewhere in kie-wb-common, because the perspective is from there.

gabriel376: thank you!
fixed

gabriel376:

ederign:
@barboras7 when you have a chance could you please review this one?

ederign:
The war is built here: https://github.com/kiegroup/kie-wb-common/pull/2986

barboras7:
@gabriel376,
from the description of the task: 

> 3. Add a "View deployment details" menu item to the Deploy drop-down on the project page. The menu item is only enabled once the user has deployed the project. Selecting this menu item also displays the Execution Servers page with no default selection.

This does not work for me. Even for new projects, I still see the "View deployment details" option as clickable. Also, I see this option for projects when I run the business central without kie-server (therefore no way to deploy the projects). I think the item should also be disabled for this case.

gabriel376:
Hi @barboras7,

You are reading the old Jira description.
There was a long discussion about the visibility of this button on Jira comments and the Jira description was then updated.

Here is the new description:
> 3. Add a "View deployment details" menu item to the Deploy drop-down on the project page. The menu item is always visible and the button always enabled. Selecting this menu item also displays the Execution Servers page with no default selection. View Deployment Details button will only be visible for users that has access to the Execution Server Perspective.

barboras7:
@gabriel376, 
I'm sorry, I did not notice that. The description should also be updated in the corresponding BAPL then. 
The PRs look good, approving.

PR #830, opened 2019-11-01
[AF-2305] Keeping asset locks if the user decides to stay on the page

caponetto:
**Task**: [AF-2305](https://issues.jboss.org/browse/AF-2305)

**Cause**:

The root cause of this bug is that `LockManager` also has a `WindowClosingHandler`, responsible for releasing asset locks when users close the browser.
Due to [AF-2214](https://issues.jboss.org/browse/AF-2214), `WindowClosingHandler` does not guarantee anymore that users have actually closed the browser.
Thus, users lose the lock when they decide to stay on the page.

As in [AF-2214](https://issues.jboss.org/browse/AF-2214), `WindowClosingHandler` can be replaced by the `ClosedHandler`.
Then, the lock is released only when the user actually decides to close the browser.
However, there is one limitation:
In Firefox and MS Edge, the event is fired when a _tab_ is closed, but it is **not** fired when the _browser_ is closed.

**Countermeasure**:

 [AF-2214](https://issues.jboss.org/browse/AF-2214) will be applied only for chrome browsers.

**Expectation**:

Chrome browsers:
The browser will ask the user for confirmation when closing the tab, closing the browser or refreshing the page when there are unsaved changes.
The asset lock is kept if the user decides to stay on the page.
The asset lock is released if the user decides to close the browser.

DEMO
AF-2305-chrome

Other browsers:
Basically, the behavior before [AF-2214](https://issues.jboss.org/browse/AF-2214) has been restored.
There will be no warning for users when closing the browser even if they have unsaved changes.
The asset lock is released when the browser is closed.

DEMO
AF-2305-ff

### DIFF
--- a/uberfire-backend/uberfire-backend-server/pom.xml
+++ b/uberfire-backend/uberfire-backend-server/pom.xml
@@ -64,6 +64,11 @@
       <artifactId>uberfire-ssh-api</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>org.uberfire</groupId>
+      <artifactId>uberfire-server</artifactId>
+    </dependency>
+
     <!-- Formatting -->
     <dependency>
       <groupId>org.ocpsoft.prettytime</groupId>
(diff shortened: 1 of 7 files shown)


paulovmr:

PR #835, opened 2019-11-07
[AF-2322] Disable X on Notifications when AutoHide is set

gabriel376:
[[AF-2322] Disable X on Notifications when AutoHide is set](https://issues.jboss.org/browse/AF-2322)

Branch was built on top of https://github.com/kiegroup/appformer/pull/829, so just last commit (89061323eb84f2fc0755b05ae67bd634f3164e16) apply to this PR.

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/notifications/NotificationPopupView.java
+++ b/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/notifications/NotificationPopupView.java
@@ -113,6 +113,8 @@ public class NotificationPopupView extends DecoratedPopupPanel {
             notification.addCloseHandler(evt -> onCompleteCommand.execute());
         }
 
+        notification.setDismissable(!autoHide);
+
         s.run();
     }
 


barboras7:

barboras7:
Hi @gabriel376, @ederign, @paulovmr,
although I understand the motivation for this fix, let me share some concerns I see here. 

**User point of view:**
The notifications might get annoying when doing some actions. Let me share an example. Imagine an experienced user working with a DRL file in the DRL editor. The user wants to validate some changes and if the validation is successful, they want to save the changes they just made. After validation, the notification shows up. Then, to be able to save the changes, the user has to wait until the notification disappears. I know that this might sound as a small issue but I believe this might get annoying when doing this frequently.

**Testing point of view:**
This is a big concern. Let me explain how our Selenium tests work. If you do an action that makes the notification appear, Selenium checks if the notification appeared and closes it with the "X" button. The reason for this is quite obvious - if the notification is present and we want to click some button, it might be present although not clickable for Selenium as it is hidden by the notification. 
The problem that we would face here is the need to wait every time for the notification to disappear - after creating a project, creating an asset, building a project, etc. That would cost us lot of time in tests because the notifications are present everywhere.

ederign:
Hi @barboras7 from user point of view this decision was made by UX team after a long round of discussions: https://issues.jboss.org/browse/AF-2322 and if we keep on this direction unfortunately we will have to wait for the notification disappears on our selenium test suit. (/cc @porcelli @amysueg)

amysueg:
@gabriel376 @ederign @barboras7 @porcelli @lclay2 
The UX team met to discuss the concerns raised in this PR by @barboras7 
We agreed to the following change:
Include the "X" on the right side of all notifications.
For notifications without a link included, allow the user to dismiss the notification either after a timed interval or when the user selects the "X".
For notifications that include a link, only dismiss by user action - either selecting the link or the "X" button.

paulovmr:
Closing this since the behavior agreed by UX team is already on master.

PR #843, opened 2019-11-12
DROOLS-4735: [DMN Designer] Grid performance is dire

manstis:
See https://issues.jboss.org/browse/DROOLS-4735

Part of an ensemble:
- https://github.com/kiegroup/appformer/pull/843
- https://github.com/kiegroup/kie-wb-common/pull/3004
- https://github.com/kiegroup/drools-wb/pull/1264

This PR really moves calculation of row heights from "on demand" to be cached in the `RenderingInformation` class. Subsequent uses of rows' heights was then moved to use the cached value if a `RenderingInformation` instance was available (i.e had been instantiated elsewhere in the class and could be passed as a parameter etc to needy code).

### DIFF
--- a/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/pom.xml
+++ b/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/pom.xml
@@ -44,6 +44,12 @@
       </exclusions>
     </dependency>
 
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>testing</scope>
+    </dependency>
+
   </dependencies>
 
 </project>
--- a/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/pom.xml
+++ b/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/pom.xml
@@ -47,7 +47,7 @@
     <dependency>
       <groupId>org.assertj</groupId>
       <artifactId>assertj-core</artifactId>
-      <scope>testing</scope>
+      <scope>test</scope>
     </dependency>
 
   </dependencies>
--- a/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseDownHandler.java
+++ b/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseDownHandler.java
@@ -158,11 +158,13 @@ public class GridWidgetDnDMouseDownHandler implements NodeMouseDownHandler {
 
         final Bounds bounds = renderingInformation.getBounds();
         final GridRow row = activeGridRows.get(0);
-        final double rowOffsetY = rendererHelper.getRowOffset(row) + view.getRenderer().getHeaderHeight();
+        final int rowIndex = view.getModel().getRows().indexOf(row);
+        final List<Double> allRowHeights = renderingInformation.getAllRowHeights();
+        final double rowOffsetY = rendererHelper.getRowOffset(rowIndex, allRowHeights) + view.getRenderer().getHeaderHeight();
 
         final double highlightWidth = Math.min(bounds.getX() + bounds.getWidth() - view.getComputedLocation().getX(),
                                                view.getWidth());
-        final double highlightHeight = row.getHeight();
+        final double highlightHeight = allRowHeights.get(rowIndex);
 
         state.getEventColumnHighlight().setWidth(highlightWidth)
                 .setHeight(highlightHeight)
--- a/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseDownHandler.java
+++ b/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseDownHandler.java
@@ -124,10 +124,11 @@ public class GridWidgetDnDMouseDownHandler implements NodeMouseDownHandler {
                                                           view,
                                                           headerMinY);
 
-        state.getEventColumnHighlight().setWidth(highlightWidth)
-                .setHeight(highlightHeight)
-                .setX(view.getComputedLocation().getX() + activeColumnX)
-                .setY(view.getComputedLocation().getY() + headerMinY);
+        final GridWidgetDnDProxy highlight = state.getEventColumnHighlight();
+        highlight.setWidth(highlightWidth);
+        highlight.setHeight(highlightHeight);
+        highlight.setX(view.getComputedLocation().getX() + activeColumnX);
+        highlight.setY(view.getComputedLocation().getY() + headerMinY);
         layer.add(state.getEventColumnHighlight());
         layer.getLayer().batch();
     }
@@ -166,10 +167,11 @@ public class GridWidgetDnDMouseDownHandler implements NodeMouseDownHandler {
                                                view.getWidth());
         final double highlightHeight = allRowHeights.get(rowIndex);
 
-        state.getEventColumnHighlight().setWidth(highlightWidth)
-                .setHeight(highlightHeight)
-                .setX(view.getComputedLocation().getX())
-                .setY(view.getComputedLocation().getY() + rowOffsetY);
+        final GridWidgetDnDProxy highlight = state.getEventColumnHighlight();
+        highlight.setWidth(highlightWidth);
+        highlight.setHeight(highlightHeight);
+        highlight.setX(view.getComputedLocation().getX());
+        highlight.setY(view.getComputedLocation().getY() + rowOffsetY);
         layer.add(state.getEventColumnHighlight());
         layer.getLayer().batch();
     }
--- a/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseMoveHandler.java
+++ b/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseMoveHandler.java
@@ -139,6 +139,7 @@ public class GridWidgetDnDMouseMoveHandler implements NodeMouseMoveHandler {
             if (cy < headerMaxY) {
                 //Check for column moving
                 findMovableColumns(gridWidget,
+                                   renderingInformation,
                                    headerHeight - headerRowsYOffset,
                                    headerMinY,
                                    cx,
@@ -146,11 +147,13 @@ public class GridWidgetDnDMouseMoveHandler implements NodeMouseMoveHandler {
             } else {
                 //Check for movable rows
                 findMovableRows(gridWidget,
+                                renderingInformation,
                                 cx,
                                 cy);
 
                 //Check for column resizing
                 findResizableColumn(gridWidget,
+                                    renderingInformation,
                                     cx);
             }
 
@@ -168,7 +171,7 @@ public class GridWidgetDnDMouseMoveHandler implements NodeMouseMoveHandler {
         }
     }
 
-    private void setCursor(final Style.Cursor cursor) {
+    protected void setCursor(final Style.Cursor cursor) {
         for (IMediator mediator : layer.getViewport().getMediators()) {
             if (mediator instanceof RestrictedMousePanMediator) {
                 if (((RestrictedMousePanMediator) mediator).isDragging()) {
@@ -181,10 +184,9 @@ public class GridWidgetDnDMouseMoveHandler implements NodeMouseMoveHandler {
     }
 
     protected void findResizableColumn(final GridWidget view,
+                                       final BaseGridRendererHelper.RenderingInformation renderingInformation,
                                        final double cx) {
         //Gather information on columns
-        final BaseGridRendererHelper rendererHelper = view.getRendererHelper();
-        final BaseGridRendererHelper.RenderingInformation renderingInformation = rendererHelper.getRenderingInformation();
         if (renderingInformation == null) {
             return;
         }
@@ -243,13 +245,12 @@ public class GridWidgetDnDMouseMoveHandler implements NodeMouseMoveHandler {
     }
 
     protected void findMovableColumns(final GridWidget view,
+                                      final BaseGridRendererHelper.RenderingInformation renderingInformation,
                                       final double headerRowsHeight,
                                       final double headerMinY,
                                       final double cx,
                                       final double cy) {
         //Gather information on columns
-        final BaseGridRendererHelper rendererHelper = view.getRendererHelper();
-        final BaseGridRendererHelper.RenderingInformation renderingInformation = rendererHelper.getRenderingInformation();
         if (renderingInformation == null) {
             return;
         }
@@ -327,25 +328,28 @@ public class GridWidgetDnDMouseMoveHandler implements NodeMouseMoveHandler {
     }
 
     protected void findMovableRows(final GridWidget view,
+                                   final BaseGridRendererHelper.RenderingInformation renderingInformation,
                                    final double cx,
                                    final double cy) {
         if (!isOverRowDragHandleColumn(view,
+                                       renderingInformation,
                                        cx)) {
             return;
         }
 
         final GridData gridModel = view.getModel();
         final GridRenderer renderer = view.getRenderer();
+        final List<Double> allRowHeights = renderingInformation.getAllRowHeights();
 
         //Get row index
         if (gridModel.getRowCount() == 0) {
             return;
         }
-        GridRow row;
+        double rowHeight;
         int uiRowIndex = 0;
         double offsetY = cy - renderer.getHeaderHeight();
-        while ((row = gridModel.getRow(uiRowIndex)).getHeight() < offsetY) {
-            offsetY = offsetY - row.getHeight();
+        while ((rowHeight = allRowHeights.get(uiRowIndex)) < offsetY) {
+            offsetY = offsetY - rowHeight;
             uiRowIndex++;
         }
         if (uiRowIndex < 0 || uiRowIndex > gridModel.getRowCount() - 1) {
@@ -353,7 +357,7 @@ public class GridWidgetDnDMouseMoveHandler implements NodeMouseMoveHandler {
         }
 
         //Add row over which MouseEvent occurred
-        final List<GridRow> rows = new ArrayList<GridRow>();
+        final List<GridRow> rows = new ArrayList<>();
         rows.add(gridModel.getRow(uiRowIndex));
 
         //Add any other collapsed rows
@@ -370,10 +374,9 @@ public class GridWidgetDnDMouseMoveHandler implements NodeMouseMoveHandler {
     }
 
     private boolean isOverRowDragHandleColumn(final GridWidget view,
+                                              final BaseGridRendererHelper.RenderingInformation renderingInformation,
                                               final double cx) {
         //Gather information on columns
-        final BaseGridRendererHelper rendererHelper = view.getRendererHelper();
-        final BaseGridRendererHelper.RenderingInformation renderingInformation = rendererHelper.getRenderingInformation();
         if (renderingInformation == null) {
             return false;
         }
@@ -596,7 +599,9 @@ public class GridWidgetDnDMouseMoveHandler implements NodeMouseMoveHandler {
         final List<GridColumn<?>> allGridColumns = activeGridModel.getColumns();
 
         final BaseGridRendererHelper rendererHelper = activeGridWidget.getRendererHelper();
+        final BaseGridRendererHelper.RenderingInformation renderingInformation = rendererHelper.getRenderingInformation();
         final GridRenderer renderer = activeGridWidget.getRenderer();
+        final List<Double> allRowHeights = renderingInformation.getAllRowHeights();
         final double headerHeight = renderer.getHeaderHeight();
 
         final GridRow leadRow = activeGridRows.get(0);
@@ -611,11 +616,11 @@ public class GridWidgetDnDMouseMoveHandler implements NodeMouseMoveHandler {
         }
 
         //Find new row index
-        GridRow row;
+        double rowHeight;
         int uiRowIndex = 0;
         double offsetY = cy - headerHeight;
-        while ((row = activeGridModel.getRow(uiRowIndex)).getHeight() < offsetY) {
-            offsetY = offsetY - row.getHeight();
+        while ((rowHeight = allRowHeights.get(uiRowIndex)) < offsetY) {
+            offsetY = offsetY - rowHeight;
             uiRowIndex++;
         }
         if (uiRowIndex < 0 || uiRowIndex > activeGridModel.getRowCount() - 1) {
@@ -627,12 +632,12 @@ public class GridWidgetDnDMouseMoveHandler implements NodeMouseMoveHandler {
             return;
         } else if (uiRowIndex < activeGridModel.getRows().indexOf(leadRow)) {
             //Don't move up if the pointer is in the bottom half of the target row.
-            if (offsetY > activeGridModel.getRow(uiRowIndex).getHeight() / 2) {
+            if (offsetY > allRowHeights.get(uiRowIndex) / 2) {
                 return;
             }
         } else if (uiRowIndex > activeGridModel.getRows().indexOf(leadRow)) {
             //Don't move down if the pointer is in the top half of the target row.
-            if (offsetY < activeGridModel.getRow(uiRowIndex).getHeight() / 2) {
+            if (offsetY < allRowHeights.get(uiRowIndex) / 2) {
                 return;
             }
         }
--- a/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseMoveHandler.java
+++ b/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseMoveHandler.java
@@ -597,9 +597,7 @@ public class GridWidgetDnDMouseMoveHandler implements NodeMouseMoveHandler {
         final List<GridColumn<?>> allGridColumns = activeGridModel.getColumns();
 
         final BaseGridRendererHelper rendererHelper = activeGridWidget.getRendererHelper();
-        final BaseGridRendererHelper.RenderingInformation renderingInformation = rendererHelper.getRenderingInformation();
         final GridRenderer renderer = activeGridWidget.getRenderer();
-        final List<Double> allRowHeights = renderingInformation.getAllRowHeights();
         final double headerHeight = renderer.getHeaderHeight();
 
         final GridRow leadRow = activeGridRows.get(0);
@@ -614,11 +612,11 @@ public class GridWidgetDnDMouseMoveHandler implements NodeMouseMoveHandler {
         }
 
         //Find new row index
-        double rowHeight;
+        GridRow row;
         int uiRowIndex = 0;
         double offsetY = cy - headerHeight;
-        while ((rowHeight = allRowHeights.get(uiRowIndex)) < offsetY) {
-            offsetY = offsetY - rowHeight;
+        while ((row = activeGridModel.getRow(uiRowIndex)).getHeight() < offsetY) {
+            offsetY = offsetY - row.getHeight();
             uiRowIndex++;
         }
         if (uiRowIndex < 0 || uiRowIndex > activeGridModel.getRowCount() - 1) {
@@ -630,12 +628,12 @@ public class GridWidgetDnDMouseMoveHandler implements NodeMouseMoveHandler {
             return;
         } else if (uiRowIndex < activeGridModel.getRows().indexOf(leadRow)) {
             //Don't move up if the pointer is in the bottom half of the target row.
-            if (offsetY > allRowHeights.get(uiRowIndex) / 2) {
+            if (offsetY > activeGridModel.getRow(uiRowIndex).getHeight() / 2) {
                 return;
             }
         } else if (uiRowIndex > activeGridModel.getRows().indexOf(leadRow)) {
             //Don't move down if the pointer is in the top half of the target row.
-            if (offsetY < allRowHeights.get(uiRowIndex) / 2) {
+            if (offsetY < activeGridModel.getRow(uiRowIndex).getHeight() / 2) {
                 return;
             }
         }
--- a/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseMoveHandler.java
+++ b/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseMoveHandler.java
@@ -331,8 +331,7 @@ public class GridWidgetDnDMouseMoveHandler implements NodeMouseMoveHandler {
                                    final BaseGridRendererHelper.RenderingInformation renderingInformation,
                                    final double cx,
                                    final double cy) {
-        if (!isOverRowDragHandleColumn(view,
-                                       renderingInformation,
+        if (!isOverRowDragHandleColumn(renderingInformation,
                                        cx)) {
             return;
         }
@@ -373,8 +372,7 @@ public class GridWidgetDnDMouseMoveHandler implements NodeMouseMoveHandler {
         setCursor(Style.Cursor.MOVE);
     }
 
-    private boolean isOverRowDragHandleColumn(final GridWidget view,
-                                              final BaseGridRendererHelper.RenderingInformation renderingInformation,
+    private boolean isOverRowDragHandleColumn(final BaseGridRendererHelper.RenderingInformation renderingInformation,
                                               final double cx) {
         //Gather information on columns
         if (renderingInformation == null) {
--- a/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/impl/BaseGridWidget.java
+++ b/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/impl/BaseGridWidget.java
@@ -24,7 +24,6 @@ import java.util.Map;
 import java.util.Queue;
 import java.util.function.BiFunction;
 import java.util.function.Function;
-import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
@@ -59,6 +58,8 @@ import org.uberfire.ext.wires.core.grids.client.widget.grid.selections.impl.Base
 import org.uberfire.ext.wires.core.grids.client.widget.layer.GridSelectionManager;
 import org.uberfire.ext.wires.core.grids.client.widget.layer.impl.DefaultGridLayer;
 import org.uberfire.ext.wires.core.grids.client.widget.layer.pinning.GridPinnedModeManager;
+
+import static org.uberfire.ext.wires.core.grids.client.util.Logging.log;
 
 /**
  * The base of all GridWidgets.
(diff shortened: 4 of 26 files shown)


manstis:
Much of the "[new untested code](https://sonarcloud.io/component_measures?id=org.uberfire%3Auberfire-parent&metric=new_coverage&pullRequest=843&selected=org.uberfire%3Auberfire-parent%3Auberfire-extensions%2Fuberfire-wires%2Fuberfire-wires-core%2Fuberfire-wires-core-grids%2Fsrc%2Fmain%2Fjava%2Forg%2Fuberfire%2Fext%2Fwires%2Fcore%2Fgrids%2Fclient%2Fwidget%2Fgrid%2Fimpl%2FBaseGridWidget.java)" in `BaseGridWidget` is the logging I added to gain performance metrics to help identify where the slow bits were.. we can either ignore Sonar's complaint or I can remove the logging.

manstis:
Jenkins please execute full downstream build.

manstis:
Sonar's complaints about insufficient test coverage:
- [BaseGridRendererHelper](https://sonarcloud.io/component_measures?id=org.uberfire%3Auberfire-parent&metric=new_coverage&pullRequest=843&selected=org.uberfire%3Auberfire-parent%3Auberfire-extensions%2Fuberfire-wires%2Fuberfire-wires-core%2Fuberfire-wires-core-grids%2Fsrc%2Fmain%2Fjava%2Forg%2Fuberfire%2Fext%2Fwires%2Fcore%2Fgrids%2Fclient%2Fwidget%2Fgrid%2Frenderers%2Fgrids%2Fimpl%2FBaseGridRendererHelper.java)
- [BaseGridWidget](https://sonarcloud.io/component_measures?id=org.uberfire%3Auberfire-parent&metric=new_coverage&pullRequest=843&selected=org.uberfire%3Auberfire-parent%3Auberfire-extensions%2Fuberfire-wires%2Fuberfire-wires-core%2Fuberfire-wires-core-grids%2Fsrc%2Fmain%2Fjava%2Forg%2Fuberfire%2Fext%2Fwires%2Fcore%2Fgrids%2Fclient%2Fwidget%2Fgrid%2Fimpl%2FBaseGridWidget.java)

The new, untested lines are pretty much all for logging I added to trace back performance issues. The `if(..)` statements were added for another Sonar rule that was violated (use of `String` concatenation) and limitations in GWT's emulation of Java libraries.. so my options are to ignore Sonar's complaint - look manually at the 2 files for coverage; or delete logging (may be useful again in the future).

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/impl/BaseGridWidget.java
```diff
@@ -244,6 +248,12 @@ public double getHeight() {
         return height;
```
danielezonca: Should `getHeight()` now be deprecated?

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/renderers/columns/impl/ColumnRenderingStrategyMerged.java
```diff
@@ -115,7 +115,8 @@
         //Column content
         commands.add((GridRenderer.RenderBodyGridContentCommand) (rc) -> {
             if (columnRenderingConstraint.apply(rc.isSelectionLayer(), column)) {
-                final Group columnGroup = new Group().setX(x);
+                final Group columnGroup = GWT.create(Group.class);
```
danielezonca: Is `GWT.create` better in terms of performance compared to `new`?

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/renderers/columns/impl/ColumnRenderingStrategyMerged.java
```diff
@@ -305,10 +306,10 @@ protected static boolean isCollapsedCellMixedValue(final GridData model,
     }
 
     protected static double getCellHeight(final int rowIndex,
-                                          final GridData model,
+                                          final List<Double> allRowHeights,
                                           final GridCell<?> cell) {
         return IntStream.range(rowIndex, rowIndex + cell.getMergedCellCount())
-                .mapToDouble(index -> model.getRow(index).getHeight())
+                .mapToDouble(allRowHeights::get)
                 .sum();
```
danielezonca: Even here a foreach should be better

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseMoveHandler.java
```diff
@@ -181,10 +184,9 @@ private void setCursor(final Style.Cursor cursor) {
     }
 
     protected void findResizableColumn(final GridWidget view,
+                                       final BaseGridRendererHelper.RenderingInformation renderingInformation,
                                        final double cx) {
         //Gather information on columns
-        final BaseGridRendererHelper rendererHelper = view.getRendererHelper();
-        final BaseGridRendererHelper.RenderingInformation renderingInformation = rendererHelper.getRenderingInformation();
```
danielezonca: Is this `renderingInformation` different from the one now obtained as parameter? If it is different I think we should remove `getRenderingInformation()` from `BaseGridRendererHelper` because it is not reliable (and maybe `getRendererHelper()` from `view` too?)

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseMoveHandler.java
```diff
@@ -596,7 +597,9 @@ protected void handleRowMove(final NodeMouseMoveEvent event) {
         final List<GridColumn<?>> allGridColumns = activeGridModel.getColumns();
 
         final BaseGridRendererHelper rendererHelper = activeGridWidget.getRendererHelper();
+        final BaseGridRendererHelper.RenderingInformation renderingInformation = rendererHelper.getRenderingInformation();
```
danielezonca: Mmmm ok so here it seems that `rendererHelper.getRenderingInformation()` is reliable. I'm a bit confused

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/layer/impl/GridLienzoPanel.java
```diff
@@ -123,7 +167,7 @@ void addMouseUpHandler() {
         addMouseUpHandler((e) -> refreshScrollPosition());
     }
 
-    private void setupDefaultHandlers() {
+    void setupDefaultHandlers() {
```
danielezonca: Why not `protected`?

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/layer/impl/GridLienzoPanel.java
```diff
@@ -83,6 +83,38 @@ public void onResize() {
                         height);
     }
 
+    public GridLienzoPanel(final DefaultGridLayer defaultGridLayer) {
+        this(new LienzoPanel() {
+                 @Override
+                 public void onResize() {
```
danielezonca: If I remember correctly in Test Scenario we use a `onResize()` to force rendering so I don't know if this could break that (not sure it is the same `onResize()` method).
@gitgabrio ^^

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/renderers/grids/impl/BaseGridRendererHelper.java
```diff
@@ -138,9 +158,18 @@ public double getWidth(final List<GridColumn<?>> columns) {
...
-        final List<GridColumn<?>> floatingColumns = new ArrayList<GridColumn<?>>();
+        final List<GridColumn<?>> allColumns = new ArrayList<>();
+        final List<GridColumn<?>> bodyColumns = new ArrayList<>();
+        final List<GridColumn<?>> floatingColumns = new ArrayList<>();
+
+        double viewHeight = 0;
+        final int rowCount = model.getRowCount();
+        final List<Double> rowHeights = new ArrayList<>();
+        for (GridRow row : model.getRows()) {
+            final double rowHeight = row.getHeight();
```
danielezonca: Is this `getHeight()` in the same slowpath?

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/impl/BaseGridWidget.java
```diff
@@ -244,6 +248,12 @@ public double getHeight() {
...
     }
 
+    private double getHeight(final BaseGridRendererHelper.RenderingInformation renderingInformation) {
+        double height = renderer.getHeaderHeight();
+        height = height + renderingInformation.getAllRowHeights().stream().reduce(0d, Double::sum);
```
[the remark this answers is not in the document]
gitgabrio: @manstis 
or - if feasible - parallel stream

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/renderers/columns/impl/ColumnRenderingStrategyMerged.java
gitgabrio: @danielezonca 
`GWT.create` is not equivalent to `new`

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/layer/impl/GridLienzoPanel.java
gitgabrio: @manstis @danielezonca 
Yes, we call `onResize` to force redrwan of the grid, IIRC

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/impl/BaseGridWidget.java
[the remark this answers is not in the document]
danielezonca: From what I can see `parallel()` in GWT is just a noop method https://github.com/gwtproject/gwt/blob/master/user/super/com/google/gwt/emul/java/util/stream/StreamImpl.java#L284-L287

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/pom.xml
```diff
@@ -44,6 +44,12 @@
       </exclusions>
     </dependency>
 
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>testing</scope>
```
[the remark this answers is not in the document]
manstis: Fixed. Thanks.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseDownHandler.java
```diff
@@ -124,10 +124,11 @@ protected void showColumnHighlight(final GridWidget view,
...
 
-        state.getEventColumnHighlight().setWidth(highlightWidth)
-                .setHeight(highlightHeight)
-                .setX(view.getComputedLocation().getX() + activeColumnX)
-                .setY(view.getComputedLocation().getY() + headerMinY);
+        final GridWidgetDnDProxy highlight = state.getEventColumnHighlight();
+        highlight.setWidth(highlightWidth);
+        highlight.setHeight(highlightHeight);
+        highlight.setX(view.getComputedLocation().getX() + activeColumnX);
+        highlight.setY(view.getComputedLocation().getY() + headerMinY);
```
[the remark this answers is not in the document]
manstis: To ease Unit Testing as I don't need to mock the return of the setter methods when `x().y().z()` are _chained_ as opposed to `x(); y(); z();`

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseMoveHandler.java
```diff
@@ -139,18 +139,21 @@ protected void findGridColumn(final NodeMouseMoveEvent event) {
...
                 findMovableRows(gridWidget,
+                                renderingInformation,
                                 cx,
                                 cy);
 
                 //Check for column resizing
                 findResizableColumn(gridWidget,
+                                    renderingInformation,
                                     cx);
             }
```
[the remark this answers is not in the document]
manstis: Yes, entirely for performance reasons. Creation of a `RenderingInformation` object is quite slow so I pass an already calculated instance to the methods as opposed to re-creating it.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/scrollbars/GridLienzoScrollHandler.java
```diff
@@ -224,27 +229,29 @@ GridLienzoPanel getPanel() {
     }
 
     DefaultGridLayer getDefaultGridLayer() {
-        return Optional.ofNullable(panel.getDefaultGridLayer()).orElse(emptyLayer());
+        //Do not use Optional.ofNullable(..).orElse(..) as the _else_ expression is *always* invoked
+        final DefaultGridLayer defaultGridLayer = panel.getDefaultGridLayer();
+        return Objects.nonNull(defaultGridLayer) ? defaultGridLayer : emptyLayer();
```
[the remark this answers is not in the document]
manstis: `Optional.ofNullable(panel.getDefaultGridLayer()).orElse(emptyLayer())` always calls `emptyLayer()` even if `panel.getDefaultGridLayer()` is non-null. Calls to `emptyLayer()` creates a new `DefaultGridLayer` which attaches event handlers in its constructor even if the object is not needed. Additional event handlers slow things down.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/test/java/org/uberfire/ext/wires/core/grids/client/util/CellContextUtilitiesTest.java
```diff
@@ -142,6 +142,7 @@ public void testMakeCellRenderContextOneRow() {
         final double headerRowsHeight = 100.0;
         final BaseGridRow row = new BaseGridRow();
         final List<GridColumn<?>> allColumns = new ArrayList<>();
+        final List<Double> allRowHeights = new ArrayList<>(Collections.singletonList(row.getHeight()));
```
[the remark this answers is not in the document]
manstis: Done. Thanks.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/test/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseDownHandlerTest.java
```diff
@@ -51,8 +62,25 @@
...
+
+    private static final double GRID_Y = 20.0;
+
+    private static final double GRID_WIDTH = 100.0;
+
+    private static final double GRID_HEIGHT = 200.0;
+
+    private static final double COLUMN_WIDTH = 25.0;
+
+    private static final double HEADER_HEIGHT = 25.0;/**/
```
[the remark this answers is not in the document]
manstis: Removed. Thanks.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/test/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseDownHandlerTest.java
```diff
@@ -192,4 +268,64 @@ public void stateGridMovePendingMovesToGridMove() {
...
+    public void testColumnHighlight() {
+        final GridWidgetDnDProxy highlight = mock(GridWidgetDnDProxy.class);
+        final Bounds bounds = new BaseBounds(GRID_X, GRID_Y, GRID_WIDTH, GRID_HEIGHT);
+
+        when(state.getEventColumnHighlight()).thenReturn(highlight);
+        when(state.getActiveGridWidget()).thenReturn(gridWidget);
+        when(state.getOperation()).thenReturn(GridWidgetHandlersOperation.COLUMN_MOVE_PENDING);
+        when(state.getActiveGridColumns()).thenReturn(new ArrayList<GridColumn<?>>() {{
+            add(uiColumn);
+        }});
```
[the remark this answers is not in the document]
manstis: Updated. Thanks.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/test/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseDownHandlerTest.java
```diff
@@ -192,4 +268,64 @@ public void stateGridMovePendingMovesToGridMove() {
...
+        final Bounds bounds = new BaseBounds(GRID_X, GRID_Y, GRID_WIDTH, GRID_HEIGHT);
+        final GridData gridData = new BaseGridData();
+        gridData.appendRow(uiRow);
+
+        when(state.getEventColumnHighlight()).thenReturn(highlight);
+        when(state.getActiveGridWidget()).thenReturn(gridWidget);
+        when(state.getOperation()).thenReturn(GridWidgetHandlersOperation.ROW_MOVE_PENDING);
+        when(state.getActiveGridRows()).thenReturn(new ArrayList<GridRow>() {{
+            add(uiRow);
+        }});
```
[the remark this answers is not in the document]
manstis: Updated. Thanks.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/test/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseMoveHandlerTest.java
```diff
@@ -481,7 +506,43 @@ public void findMovableRows() {
...
+    @Test
+    public void handleRowMove() {
+        final List<GridRow> existingRowOrder = new ArrayList<>(uiModel.getRows());
+        final GridWidgetDnDProxy highlight = mock(GridWidgetDnDProxy.class);
+        when(state.getEventColumnHighlight()).thenReturn(highlight);
+        when(state.getOperation()).thenReturn(GridWidgetHandlersOperation.NONE);
+        when(gridWidget.isVisible()).thenReturn(true);
+        when(layer.getGridWidgets()).thenReturn(new HashSet<GridWidget>() {{
+            add(gridWidget);
+        }});
```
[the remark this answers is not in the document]
manstis: Updated (and all other occurrences in this class). Thanks.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/test/java/org/uberfire/ext/wires/core/grids/client/util/CellContextUtilitiesTest.java
```diff
@@ -178,18 +180,21 @@ public void testMakeCellRenderContextThreeRows() {
         final BaseGridRow row2 = new BaseGridRow();
         final BaseGridRow row3 = new BaseGridRow();
         final List<GridColumn<?>> allColumns = new ArrayList<>();
+        final List<Double> allRowHeights = new ArrayList<>(Collections.nCopies(3, row1.getHeight()));
```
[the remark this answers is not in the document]
manstis: This test checks the `GridBodyCellEditContext` for the 3rd row (of 3). Consequentially this collection needs to contain row heights of all 3 rows; hence use of `Collections.nCopies(3,.....)`

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/impl/BaseGridWidget.java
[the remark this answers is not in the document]
manstis: Changed to use a basic loop. 

As @danielezonca rightly says `parallel()` has no effect in the land of GWT.. plus a basic loop to sum some doubles is arguably quicker (without profiling) than the hoops and loops a parallel stream would need go through.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/renderers/columns/impl/ColumnRenderingStrategyMerged.java
manstis: GWT.create(..) allows `GwtMockitoTestRunner` and `LienzoMockitoTestRunner` to provide mocks in Unit Tests. Use of `new` does not.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/impl/BaseGridWidget.java
manstis: No, a grid still has height, as do rows and now cells. A row's height is that of it's tallest cell and a grids height is the sum of its rows heights.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseMoveHandler.java
manstis: No it is exactly the same (given the same state of the grid, position, number of rows, columns etc). All calls to `BaseGridRendererHelper.getRenderingInformation(..)` will return identical results (although the result is is not cached and is re-created on each invocation). So passing an existing instance around removes the need to re-create it over and over again.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/renderers/columns/impl/ColumnRenderingStrategyMerged.java
manstis: Updated.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseMoveHandler.java
manstis: I changed this to get all rows heights once and re-use the collection in the remainder of the method. However on review getting ALL rows' heights (all be it once) is probably slower than getting the height of the rows involved in the drag operation (which, since is only visible rows, is less for large tables). I have therefore reverted the change to this method.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/layer/impl/GridLienzoPanel.java
manstis: This class contains lots of default access level methods. It is a favourite approach by @karreiro for unit testing where a method needs mocking but not part of the _public_ API. I took this opportunity to make this method the same access level as other _helper_ methods on the class.. I see @danielezonca is an advocate of not using this access level. I see @karreiro is an advocate... There's only one way to resolve this...... FIGHT!

No, seriously, I'm open to understanding both of your view points. I think IF a method is considered something that is considered purely a 100% internal implementation detail, default access has it's merits. However who is to know what methods a sub-class developer might like/need to override and hence use of `protected` makes sense.....

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/layer/impl/GridLienzoPanel.java
manstis: The 2 new constructors are no different to the 2 existing constructors (one of which _scesim_ would be using). Well, when I say no different, I mean their use of `LienzoPanel` is no different and they simply allow me to pass a `DefaultGridLayer`. I see this to have no affect on _scesim_.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/renderers/grids/impl/BaseGridRendererHelper.java
manstis: No really. This use is in the `getRenderingInformation()` method that gets the _all row heights_ collection used elsewhere (to avoid this same loop being called again and again elsewhere)... So, yes, it is the same loop as has been replaced but now it should be the only loop of this nature!

manstis:
@danielezonca @gitgabrio @jomarko All updates completed, as applicable.

Please re-review at your convenience (and the other 2 please). Thanks.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/impl/BaseGridWidget.java
```diff
@@ -313,9 +334,11 @@ protected void drawWithoutTransforms(Context2D context,
         }
 
         //Then render to the canvas
+        currentTimeMillis = log(" - Pre- super.drawWithoutTransforms()");
         super.drawWithoutTransforms(context,
                                     alpha,
                                     bb);
+        log(" - Post- super.drawWithoutTransforms() - " + (System.currentTimeMillis() - currentTimeMillis) + "ms");
```
gitgabrio: log(" - Post- super.drawWithoutTransforms() - ", currentTimeMillis);

 private static long log(final String message, long currentTimeMillis) {
        if (LOGGER.isLoggable(Level.FINEST)) {
            LOGGER.log(Level.FINEST, message + (System.currentTimeMillis() - currentTimeMillis) + "ms");
        }
        return System.currentTimeMillis();
    }

Let's reduce boilerplate code, if possible

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/renderers/grids/impl/BaseGridRendererHelper.java
```diff
@@ -245,9 +287,13 @@ public RenderingInformation getRenderingInformation() {
                 }
             }
         }
+        if (LOGGER.isLoggable(Level.FINEST)) {
```
gitgabrio: @manstis 
Pls remove all that duplicated Logging code

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseMoveHandler.java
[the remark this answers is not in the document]
jomarko: ok, thank you

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/scrollbars/GridLienzoScrollHandler.java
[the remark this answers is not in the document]
jomarko: @manstis so you mean "always calls emptyLayer()" == it is not lazy? or else is executed even when the `Optional` value is `panel.getDefaultGridLayer`?

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/test/java/org/uberfire/ext/wires/core/grids/client/util/CellContextUtilitiesTest.java
jomarko: ok but is the `new ArrayList()` wrapper really needed?

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/layer/impl/GridLienzoPanel.java
gitgabrio: @manstis @danielezonca @karreiro 
I know Guillerme is a "default" supporter. 
Just to learn from mistakes, though, I would like to remember that when the "AbstractPopover" has been implemented, I discussed almost an hour to try to convince to use "protected" modifiers - I've been un-heard. And the first thing I had to do to re-use it in scesim has been to rewrite it because otherwise it was not re-usable.
It is extremely hard to foresee what would be the needs of a specific component, and kie-common codebase is full of widget absolutely un-reusable exactly for that reason.
If a developer "broke" a functionality overriding a method, it is his/her responsability - and this is possible regardless of the modifier applied.
In that specific case, it is almost sure that, sooner or later, someone will need to override the "default handlers" for some reason - and that should not need a widget rewrite.
A good code is something that is easy to extend/reuse, not something that works only in the specific situation where it has been firstly used.

So, let's be "pragmatic" and stop shooting on our own feet :) :) :)

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/test/java/org/uberfire/ext/wires/core/grids/client/util/CellContextUtilitiesTest.java
manstis: No. You are correct. Corrected in new commit.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/impl/BaseGridWidget.java
manstis: Changed in new commit.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/renderers/grids/impl/BaseGridRendererHelper.java
manstis: Done.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/scrollbars/GridLienzoScrollHandler.java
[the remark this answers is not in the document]
manstis: Correct!
```
public class Test {

    public static void main(String[] args) {
        System.out.println("null value-->");
        System.out.println(Optional.ofNullable(null).orElse(get()));

        System.out.println("non-null value-->");
        System.out.println(Optional.ofNullable("hello").orElse(get()));
    }

    private static String get() {
        System.out.println("orElse(..) called");
        return "default";
    }
}
```
This **always** shows `orElse(..) called` whether the value was `null` or non-`null`.

manstis:
@danielezonca @gitgabrio @jomarko All issues resolved/commented.

Please re-review at your pleasure/leisure!

manstis:
Jenkins please execute full downstream build.

manstis:
Jenkins please execute full downstream build.

manstis:
Jenkins please retest this.

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/layer/impl/GridLienzoPanel.java
manstis: Updated _default_ access to `protected` in this class.

manstis: - and I can adopt `protected` over _default_ for the reasons given ongoing.

manstis:
Jenkins execute compile downstream build

gitgabrio:
@manstis @danielezonca @yesamer  @jomarko 
Hi. Just a couple general observations. 
Changing an algorithm inside an implementation should be extremely localized inside the specific implementation. If this modification actually requires changing three repositories it means that the code is not well isolated, and it is a a sure indicator of wrong coupling between modules that should be separated.
Beside that, I still think that all the "rendering" stuff is an implementation detail that must be hidden to using developer. Said differently, as a Grid user I should not be forced to mess with the "renderers", I should simply have an API that gives me ootb sizes and positions from the Grid.
After reading this PR, I'm the more convinced that all the component would hugely benefit of a common review/rethinking, otherwise it will always be a single-man effort, and a black-box for whoever has to use it.

manstis:
@gitgabrio You can't have your cake and eat it.

You call for more methods to be protected so they can be overridden yet find umbrage when such overrides need updating when the super class changes. Is it ideal. No. Could I have made the renderer pluggable, so grids can choose to use the more efficient algorithm. Probably.

gitgabrio:
@manstis 
It is only a matter of separation between internal implementations and public API.
As a grid developer user, I would simply need an easy way to retrieve the dimensions/positions of elements - I should not have to be concerned on *how* those are calculated - and such stuff should be kept internal.
Then, if I want to change the behavior (let's say the color of the grid, the handlers of specific events, whatever) I should be able to easily do that.
Rendering the grid to the screen is a low-level detail. Improving the algorithm to make it faster is another low-level detail. Both should be kept inside the implementation and should not creep out.
There is absolutely no-sense on re-implementing them by the final user - it would mean re-implementing the component itself, since it is a *graphical* object.
What it may have sense it is to change the other behaviors. I hope to have made clear the difference. 
Please do not get it wrong, I'm only stating the obvious based on the issues we had/have on using that component, to give suggestion on how to improve it.

manstis:
Hi @gitgabrio I probably should not respond to PR comments whilst on PTO :-)

The public _API_ for the grid has not changed and this remains well defined in the various interfaces.

What did change, as you note, is the internal algorithm for rendering. This was not exposed.

The other PRs, `kie-wb-common` does not rely on the change to `uberfire-wires` but is an enhancement to just DMN to pre-calculate row heights rather than calculate them on demand. It is not necessary for other grids where the row height does not depend on cell content. The PR for `drools-wb` only fixes tests that in reality test the internal rendering algorithm (and arguably should not be in `drools-wb` per se).

Are you blocking the merge of these PRs? In which case please propose how we can remedy the impasse.

manstis:
Jenkins please execute full downstream build.

PR #854, opened 2019-11-21
AF-2353: Adding kogito-appformer-bridge

jesuino:
Merge with: 

https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1119

Related PRs:

https://github.com/kiegroup/kogito-tooling/pull/40
kiegroup/kie-wb-common#3034

### DIFF
--- a/appformer-kogito-bridge/pom.xml
+++ b/appformer-kogito-bridge/pom.xml
@@ -27,51 +27,26 @@
 
   <artifactId>appformer-kogito-bridge</artifactId>
   <packaging>jar</packaging>
+
   <dependencies>
-
-    <dependency>
-      <groupId>javax.inject</groupId>
-      <artifactId>javax.inject</artifactId>
-      <scope>provided</scope>
-    </dependency>
-
-    <dependency>
-      <groupId>org.jboss.errai</groupId>
-      <artifactId>errai-javax-enterprise</artifactId>
-    </dependency>
-
-    <dependency>
-      <groupId>org.jboss.errai</groupId>
-      <artifactId>errai-common</artifactId>
-    </dependency>
-
-    <dependency>
-      <groupId>org.jboss.errai</groupId>
-      <artifactId>errai-bus</artifactId>
-    </dependency>
-
     <dependency>
       <groupId>org.jboss.errai</groupId>
       <artifactId>errai-ioc</artifactId>
+      <scope>provided</scope>      
     </dependency>
     <dependency>
       <groupId>com.google.elemental2</groupId>
       <artifactId>elemental2-dom</artifactId>
+      <scope>provided</scope>
     </dependency>
-
     <dependency>
       <groupId>com.google.elemental2</groupId>
       <artifactId>elemental2-promise</artifactId>
+      <scope>provided</scope>
     </dependency>
-
     <dependency>
       <groupId>com.google.jsinterop</groupId>
       <artifactId>jsinterop-annotations</artifactId>
-    </dependency>
-    
-    <dependency>
-      <groupId>com.google.gwt</groupId>
-      <artifactId>gwt-user</artifactId>
       <scope>provided</scope>
     </dependency>
   </dependencies>
--- a/appformer-kogito-bridge/pom.xml
+++ b/appformer-kogito-bridge/pom.xml
@@ -1,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+       ~ Copyright 2019 Red Hat, Inc. and/or its affiliates.
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~       http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.uberfire</groupId>
+    <artifactId>uberfire-parent</artifactId>
+    <version>7.31.0-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+
+  <artifactId>appformer-kogito-bridge</artifactId>
+  <packaging>jar</packaging>
+  <dependencies>
+
+    <dependency>
+      <groupId>javax.inject</groupId>
+      <artifactId>javax.inject</artifactId>
+      <scope>provided</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.jboss.errai</groupId>
+      <artifactId>errai-javax-enterprise</artifactId>
+    </dependency>
+
+    <dependency>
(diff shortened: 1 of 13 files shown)


appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/resource/ResourceContentService.java
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package org.appformer.kogito.bridge.client.resource;
+
+import elemental2.promise.Promise;
+
+public interface ResourceContentService {
+    
+    public Promise<String> get(String uri);
```
ederign: JAvadocs.

appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/resource/producer/ResourceClientServiceProducer.java
```diff
@@ -0,0 +1,47 @@
...
+        if (isResourceContentServiceAvailable) {
+            return new ResourceContentServiceImpl();
+        } else {
+            return new MockResourceContentService();
+        }
+    }
+
+    private native boolean isResourceContentServiceAvailable()/*-{
+                                                              return typeof $wnd.envelope !== "undefined";
+                                                              }-*/;
```
tiagobento: This is probably not correctly formatted.

appformer-kogito-bridge/pom.xml
```diff
@@ -0,0 +1,78 @@
...
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.jsinterop</groupId>
+      <artifactId>jsinterop-annotations</artifactId>
+    </dependency>
+    
+    <dependency>
+      <groupId>com.google.gwt</groupId>
+      <artifactId>gwt-user</artifactId>
```
romartin: if that completely decouples us from GWT, I mean, we'll have to refactor code to inject that "resources" / "services", why does the gwt-user dependency is in scope? Could we decouple this module from GWT, also even would be nice from Errai as much as possible... does that makes sense guys?

appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/resource/ResourceContentService.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Returns a resource's content
+     * 
+     * @param uri
+     *  the resource URI relative to the workspace/project
+     * @return
+     * The resource content or null if the resource is not available
+     */
+    public Promise<String> get(String uri);
```
romartin: As the service will return the raw String content for a resource, does that makes sense considering the charset as well on the arguments, in some way? Otherwise, the javacods at lesat should specify that charset is up to the given implementaton? Does that makese sense?

appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/resource/impl/MockResourceContentService.java
```diff
@@ -0,0 +1,37 @@
...
+
+package org.appformer.kogito.bridge.client.resource.impl;
+
+import elemental2.promise.Promise;
+import org.appformer.kogito.bridge.client.resource.ResourceContentService;
+
+/**
+ * This {@link ResourceContentService} implementation is used when the envelope API is not available
+ */
+public class MockResourceContentService implements ResourceContentService {
```
romartin: Agree on having some "no-op" content service, but from my experience calling it as "Mock...Service" does probably not reflect it. Anyway, just a minor comment, just my opinion...

appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/resource/impl/ResourceContentServiceImpl.java
```diff
@@ -0,0 +1,39 @@
...
+
+import elemental2.promise.Promise;
+import org.appformer.kogito.bridge.client.resource.ResourceContentService;
+import org.appformer.kogito.bridge.client.resource.interop.ResourceContentEditorServiceWrapper;
+
+/**
+ *  A {@link ResourceContentService} implementation used when envelope API is available
+ *
+ */
+public class ResourceContentServiceImpl implements ResourceContentService {
```
romartin: Minor comment too, and similar comment as above... what does a "XXXSerivceImpl" means? What is the default "impl"?   For me the classname does not properly refelcts which / what does this impl do....

appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/resource/producer/ResourceClientServiceProducer.java
```diff
@@ -0,0 +1,47 @@
...
+    public ResourceContentService produce() {
+        boolean isResourceContentServiceAvailable = isResourceContentServiceAvailable();
+        if (isResourceContentServiceAvailable) {
+            return new ResourceContentServiceImpl();
+        } else {
+            return new MockResourceContentService();
+        }
+    }
+
+    private native boolean isResourceContentServiceAvailable()/*-{
```
romartin: I think we should prevent this as much as possible - having GWT native interfaces and code here.... hard to expose and maintain long term...

appformer-kogito-bridge/pom.xml
jesuino: Makes sense to me, I am testing the API after removing these dependencies

appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/resource/ResourceContentService.java
jesuino: I will double check with @tiagobento , but IMO it does not make a difference in the WEB ( ResourceContent is all client side) because Javascript uses UTF-16 for everything

appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/resource/impl/MockResourceContentService.java
jesuino: Thank you, will rename the class

appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/resource/producer/ResourceClientServiceProducer.java
jesuino: I completely agree, I changed to use jsinterop instead

appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/resource/impl/ResourceContentServiceImpl.java
jesuino: Changed to EnvelopeResourceContent. It won't be used directly, API users will have only to inject ResourceContentService

jesuino:
Hello Roger,

I addressed your comments on commit 270eac402d985ef0def10fd33b13d6fd17eb6157 

Let me know if you have any question.

Thanks!

appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/resource/ResourceContentService.java
romartin: ok thanks

jomarko:
Code review 01
Is it correct top level **pom.xml** doesn't reference new module?

appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/resource/interop/Envelope.java
```diff
@@ -0,0 +1,19 @@
+package org.appformer.kogito.bridge.client.resource.interop;
```
manstis: License.

jesuino:
@jomarko It was missing it in parent pom, it will be part of regular build because the kogito runtimes are!
@manstis thanks

jesuino:

jomarko:

PR #860, opened 2019-11-28
[DROOLS-4518] Extending Donut chart API to manage a label inside its hole

yesamer:
@dupliaka @gitgabrio @danielezonca @ederign 

Aim of this PR is to introduce a way to set the label inside thehole of the Donut chart provided by Dashbuilder module.

Screenshot from 2019-11-28 14-33-28

Related PR:https://github.com/kiegroup/drools-wb/pull/1267

https://issues.jboss.org/browse/DROOLS-4518

### DIFF
--- a/dashbuilder/dashbuilder-client/dashbuilder-renderers/dashbuilder-renderer-c3/src/main/java/org/dashbuilder/renderer/c3/client/charts/pie/C3PieChartDisplayer.java
+++ b/dashbuilder/dashbuilder-client/dashbuilder-renderers/dashbuilder-renderer-c3/src/main/java/org/dashbuilder/renderer/c3/client/charts/pie/C3PieChartDisplayer.java
@@ -20,12 +20,14 @@ import java.util.List;
 import javax.enterprise.context.Dependent;
 import javax.inject.Inject;
 
+import org.dashbuilder.common.client.StringUtils;
 import org.dashbuilder.common.client.widgets.FilterLabelSet;
 import org.dashbuilder.dataset.ColumnType;
 import org.dashbuilder.dataset.DataColumn;
 import org.dashbuilder.dataset.DataSetLookupConstraints;
 import org.dashbuilder.displayer.DisplayerAttributeDef;
 import org.dashbuilder.displayer.DisplayerConstraints;
+import org.dashbuilder.displayer.DisplayerSubType;
 import org.dashbuilder.renderer.c3.client.C3Displayer;
 import org.dashbuilder.renderer.c3.client.charts.CommonC3DisplayerConstants;
 import org.dashbuilder.renderer.c3.client.jsbinding.C3AxisX;
@@ -47,8 +49,11 @@ public class C3PieChartDisplayer extends C3Displayer<C3PieChartDisplayer.View> {
     @Override
     protected C3ChartConf buildConfiguration() {
         C3ChartConf conf = super.buildConfiguration();
-        C3Donut donutConf = createDonut();
-        conf.setDonut(donutConf);
+        if (DisplayerSubType.DONUT == displayerSettings.getSubtype() &&
+                !StringUtils.isBlank(displayerSettings.getDonutHoleTitle())) {
+            C3Donut donutConf = createDonut();
+            conf.setDonut(donutConf);
+        }
         return conf;
     }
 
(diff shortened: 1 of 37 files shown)


danielezonca:
jenkins execute compile downstream build

dashbuilder/dashbuilder-client/dashbuilder-renderers/dashbuilder-renderer-c3/src/main/java/org/dashbuilder/renderer/c3/client/charts/pie/C3PieChartDisplayer.java
```diff
@@ -110,4 +119,8 @@ public DisplayerConstraints createDisplayerConstraints() {
                 .supportsAttribute(DisplayerAttributeDef.SUBTYPE);
     }
 
+    private C3Donut createDonut() {
+        String donutHoleTitle = displayerSettings.getDonutHoleTitle();
```
gitgabrio: @yesamer 
Hi. This is a sort of jump of the faith - to be sure that `displayerSettings` is not null (I do not see any direct reference in this class - maybe I'm wrong) and that its "donutHoleTitle" is not null too.
Beside that - if someone want/need to customize the `C3Donut` instantiation, he/she has to override the `buildConfiguration` method to invoke a different "createDonut" (being private, `createDonut` can't be overrided)
Just general observation, though.

yesamer: @gitgabrio In my opinion, the case where `displayerSettings` is null should be considered as an error, I consider it as a mandatory information to create a chart. That parameter is used to defines the properties of a chart (eg, renderer, chart type etc). All the logic flow makes this assumption, so I think it's correct in this contexts.  
Regarding the second point, I changed the modifier as you proposed.

dashbuilder/dashbuilder-client/dashbuilder-renderers/dashbuilder-renderer-c3/src/main/java/org/dashbuilder/renderer/c3/client/jsbinding/C3Donut.java
```diff
@@ -0,0 +1,36 @@
...
+import jsinterop.annotations.JsPackage;
+import jsinterop.annotations.JsProperty;
+import jsinterop.annotations.JsType;
+
+@JsType(isNative = true, namespace = JsPackage.GLOBAL, name = "Object")
+public class C3Donut {
+
+    @JsOverlay
+    static C3Donut create(String title) {
+        C3Donut  data = new C3Donut ();
```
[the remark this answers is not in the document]
yesamer: @jesuino Done, thank you William!

dashbuilder/dashbuilder-client/dashbuilder-renderers/dashbuilder-renderer-c3/src/main/java/org/dashbuilder/renderer/c3/client/charts/pie/C3PieChartDisplayer.java
gitgabrio: @yesamer 
Yes, I'm sure that `displayerSettings` is mandatory; at the same time, in this piece of code, it is actually potentially null (not this method, nor the containing class, has any control on it) - so I think a check - with a specific exception thrown, would be nice; if not in this specific method, at least in the class. Anyway, there is no common view about that.

yesamer: @gitgabrio A null check for that field is present in the super class which contains it: 'AbstractDisplayer.draw()' : 237. And this method it's called before the chart creation and eventually before the method with highlight. 
Anyway, this not completely avoid a NPE because, as you can see, 'displayerSettings` fields are used before that `draw()` (eg `initDisplayer()` method) but this is another story....

yesamer:
Hi @jesuino ! Can you please finalize this PR review? Thanks!

jesuino:

danielezonca:
jenkins execute compile downstream build

PR #868, opened 2019-12-16
KOGITO-724: [DMN Designer] Data Types are not editable in online editors

karreiro:
See https://issues.redhat.com/browse/KOGITO-724

Before

After

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/PanelManager.java
+++ b/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/PanelManager.java
@@ -219,7 +219,7 @@ public interface PanelManager {
     void onPartMinimized(PartDefinition part);
 
     /**
-     * @return the focused {@link PanelDefinition}
+     * @return the focused {@link PartDefinition}
      */
     PartDefinition getFocusedPart();
 }
--- a/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/PanelManager.java
+++ b/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/PanelManager.java
@@ -217,4 +217,9 @@ public interface PanelManager {
      * @param part the part that has been minimized
      */
     void onPartMinimized(PartDefinition part);
+
+    /**
+     * @return the focused {@link PanelDefinition}
+     */
+    PartDefinition getFocusedPart();
 }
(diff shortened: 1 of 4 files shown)


karreiro:
Thanks for the review, @manstis :-)

Actually, `activePart` has a different lifecycle and it is nulled in the `onPartLostFocus`. While, the new `lastFocusedPart` keeps track of the last focused element and never loses its value.

Changing the logic of the `onPartFocus`, the `onPartLostFocus` and the handlers that call them is another approach (that could avoid a new field). But, personally I think that the `lastFocusedPart`-based logic is more unobtrusive.

Please let me know wdyt ;-)

Thanks again for the review.

manstis:
@karreiro Your way works fine; I was not imposing my view on you...

My way did not null-ify `activePart` as `StaticWorkbenchPanelView` would not call `panelManager.onPartLostFocus()` if the selection was for the same part so the lifecycle remained OK.. anyway.. approving.

karreiro:
@manstis Ohh! Now I understand.. (sorry for the delay). I do prefer your proposal. I'm going to update PR ;-)

karreiro:
Changes applied, @manstis.

uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/PanelManager.java
```diff
@@ -217,4 +217,9 @@ void setRoot(PerspectiveActivity activity,
      * @param part the part that has been minimized
      */
     void onPartMinimized(PartDefinition part);
+
+    /**
+     * @return the focused {@link PanelDefinition}
```
jomarko: `PartDefinition` instead of `PanelDefinition`?

manstis: Good catch!

jomarko:
@manstis I see I reported for online editors. It is enough to check `dmn-webapp-kogito-runtime` ?

manstis:
@jomarko Yes, if you `mvn clean install` the module you suggest, then unzip the WAR and double-click `index.html` you'll run the client-side version of the editor without needing to package for VSCode etc. You'll need to show the dev console and enter `gwtEditorBeans.get("DMNDiagramEditor").get().setContent("");` to show the editor.

uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/workbench/PanelManager.java
karreiro: Fixed! Thank you, @jomarko :-)

karreiro:
Changes applied, @jomarko. Thank you for the review.

ederign:
please jenkins execute full downstream build.

ederign:
@karreiro @manstis please wait a full build before merging this! (I'm just curious for get the selenium tests running)

jomarko:
I agree with full build triggering, however just a note, we have no selenium test in community that would touch DMN Designer somehow.

karreiro:
@ederign @manstis everything is green, so I think we're ready to merge! :-)

ederign:
@jomarko thanks! It's just more intended to validade appformer general safe. So we are good to go.

PR #873, opened 2020-01-03
KOGITO-652: Context for editors

jesuino:
Add a new parameter to the editor `setContent` method and make sure it is generated correctly.

Also create a wrapper object for the context. Editors can now access the context by injecting  `EditorContextProvider`.

Part of an ensemble:

https://github.com/kiegroup/kogito-tooling/pull/47
https://github.com/kiegroup/kie-wb-common/pull/3099

### DIFF
--- a/appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/context/EditorContextProvider.java
+++ b/appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/context/EditorContextProvider.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.appformer.kogito.bridge.client.context;
+
+import java.util.Optional;
+import java.util.function.Function;
+
+import javax.enterprise.context.ApplicationScoped;
+
+import org.appformer.kogito.bridge.client.context.interop.EditorContextWrapper;
+
+/**
+ * 
+ * CDI Bean to provide access to EditorContext properties
+ *
+ */
+@ApplicationScoped
+public class EditorContextProvider {
+
+    /**
+     * The channel where the editor is running
+     * @return
+     */
+    public Optional<String> getChannel() {
+        return nullSafe(wrapper -> wrapper.getChannel());
+    }
+
+    private Optional<String> nullSafe(Function<EditorContextWrapper, String> action) {
+        EditorContextWrapper wrapper = EditorContextWrapper.get();
+        if (wrapper != null) {
+            return Optional.of(action.apply(wrapper));
+        }
+        return Optional.empty();
+    }
+
+}
--- a/appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/context/EditorContextProvider.java
+++ b/appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/context/EditorContextProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -27,7 +27,7 @@ public interface EditorContextProvider {
      * Access the channel where the editor is running (e.g. ONLINE, GITHUB or VSCODE).
      * 
      * @return
-     * The channel where the editor is running or an empty optional if no channel is available.
+     * The channel where the editor is running or DEFAULT if no channel is available.
      */
     KogitoChannel getChannel();
 
--- a/appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/context/impl/EditorContextProviderImpl.java
+++ b/appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/context/impl/EditorContextProviderImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/context/impl/EditorContextProviderImpl.java
+++ b/appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/context/impl/EditorContextProviderImpl.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.appformer.kogito.bridge.client.context.impl;
+
+import javax.enterprise.context.ApplicationScoped;
+
+import org.appformer.kogito.bridge.client.context.KogitoChannel;
+import org.appformer.kogito.bridge.client.context.EditorContextProvider;
+import org.appformer.kogito.bridge.client.context.interop.EditorContextWrapper;
+import org.appformer.kogito.bridge.client.interop.WindowRef;
+
+@ApplicationScoped
+public class EditorContextProviderImpl implements EditorContextProvider {
+
+    @Override
+    public KogitoChannel getChannel() {
+        if (WindowRef.isEnvelopeAvailable()) {
+            String channel = EditorContextWrapper.get().getChannel();
+            if (channel != null) {
+                return KogitoChannel.withName(channel);
+            }
+        }
+        return KogitoChannel.DEFAULT;
+    }
+
+}
--- a/appformer-kogito-bridge/src/test/java/org/appformer/kogito/bridge/client/context/KogitoChannelTest.java
+++ b/appformer-kogito-bridge/src/test/java/org/appformer/kogito/bridge/client/context/KogitoChannelTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/appformer-kogito-bridge/src/test/java/org/appformer/kogito/bridge/client/context/KogitoChannelTest.java
+++ b/appformer-kogito-bridge/src/test/java/org/appformer/kogito/bridge/client/context/KogitoChannelTest.java
@@ -16,19 +16,20 @@
 
 package org.appformer.kogito.bridge.client.context;
 
-/**
- * 
- * Provide access to EditorContext properties
- *
- */
-public interface EditorContextProvider {
+import org.junit.Test;
 
-    /**
-     * Access the channel where the editor is running (e.g. ONLINE, GITHUB or VSCODE).
-     * 
-     * @return
-     * The channel where the editor is running or an empty optional if no channel is available.
-     */
-    KogitoChannel getChannel();
+import static org.junit.Assert.assertEquals;
+
+public class KogitoChannelTest {
+
+    @Test
+    public void withNameTest() {
+        assertEquals(KogitoChannel.GITHUB, KogitoChannel.withName("GitHub"));
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void withWrongNameTest() {
+        KogitoChannel.withName("foo");
+    }
 
 }
--- a/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/mvp/WorkbenchClientEditorActivity.java
+++ b/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/mvp/WorkbenchClientEditorActivity.java
@@ -30,9 +30,9 @@ public interface WorkbenchClientEditorActivity extends WorkbenchActivity {
      *  Set the editor content
      *  
      * @param path
-     * Content Path 
+     *  Content Relative Path
      * @param value
-     * The editor content
+     *  The editor content
      */
     void setContent(String path, String value);
 
--- a/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/mvp/WorkbenchClientEditorActivity.java
+++ b/uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/mvp/WorkbenchClientEditorActivity.java
@@ -26,11 +26,15 @@ import org.uberfire.workbench.model.ActivityResourceType;
 public interface WorkbenchClientEditorActivity extends WorkbenchActivity {
 
     /**
-     * Set the editor content
+     *  
+     *  Set the editor content
+     *  
+     * @param path
+     * Content Path 
      * @param value
-     *  The the editor content
+     * The editor content
      */
-    void setContent(String value);
+    void setContent(String path, String value);
 
     /**
      * Get the editor content
--- a/uberfire-workbench/uberfire-workbench-processors-tests/src/test/java/org/uberfire/annotations/processors/WorkbenchClientEditorProcessorTest.java
+++ b/uberfire-workbench/uberfire-workbench-processors-tests/src/test/java/org/uberfire/annotations/processors/WorkbenchClientEditorProcessorTest.java
@@ -85,7 +85,7 @@ public class WorkbenchClientEditorProcessorTest extends AbstractProcessorTest {
                                  Kind.ERROR,
                                  Diagnostic.NOPOS,
                                  Diagnostic.NOPOS,
-                                 "org.uberfire.annotations.processors.WorkbenchClientEditorTest4Activity: The WorkbenchClientEditor must provide a @SetContent annotated method that has two java.lang.String as parameters.");
+                                 "org.uberfire.annotations.processors.WorkbenchClientEditorTest4Activity: The WorkbenchClientEditor must provide a @SetContent annotated method that has two java.lang.String (path and content) as parameters.");
         assertNull(result.getActualCode());
     }
     
--- a/uberfire-workbench/uberfire-workbench-processors-tests/src/test/java/org/uberfire/annotations/processors/WorkbenchClientEditorProcessorTest.java
+++ b/uberfire-workbench/uberfire-workbench-processors-tests/src/test/java/org/uberfire/annotations/processors/WorkbenchClientEditorProcessorTest.java
@@ -85,7 +85,7 @@ public class WorkbenchClientEditorProcessorTest extends AbstractProcessorTest {
                                  Kind.ERROR,
                                  Diagnostic.NOPOS,
                                  Diagnostic.NOPOS,
-                                 "org.uberfire.annotations.processors.WorkbenchClientEditorTest4Activity: The WorkbenchClientEditor must provide a @SetContent annotated method that has a java.lang.String as parameter.");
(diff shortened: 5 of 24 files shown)


jesuino:

uberfire-workbench/uberfire-workbench-processors-tests/src/test/java/org/uberfire/annotations/processors/WorkbenchClientEditorProcessorTest.java
```diff
@@ -85,7 +85,7 @@ public void testMissingSetContent() throws FileNotFoundException {
                                  Kind.ERROR,
                                  Diagnostic.NOPOS,
                                  Diagnostic.NOPOS,
-                                 "org.uberfire.annotations.processors.WorkbenchClientEditorTest4Activity: The WorkbenchClientEditor must provide a @SetContent annotated method that has a java.lang.String as parameter.");
+                                 "org.uberfire.annotations.processors.WorkbenchClientEditorTest4Activity: The WorkbenchClientEditor must provide a @SetContent annotated method that has two java.lang.String as parameters.");
```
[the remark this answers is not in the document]
jesuino: Hello Domink, thanks for your comments. I added a small parenthesis saying that the parameters are the path and content. These annotation processors errors are threw during build, while generating the activity classes, hence I was just reusing previous messages.

uberfire-workbench/uberfire-workbench-client/src/main/java/org/uberfire/client/mvp/WorkbenchClientEditorActivity.java
```diff
@@ -26,11 +26,15 @@
 public interface WorkbenchClientEditorActivity extends WorkbenchActivity {
 
     /**
-     * Set the editor content
+     *  
+     *  Set the editor content
+     *  
+     * @param path
+     * Content Path
```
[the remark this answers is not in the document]
jesuino: Improved the comment to say that it is the relative path - in fact, full path does not make sense in GITHUB channel, it would make sense in VSCODE and in the future DESKTOP channel (full path of a file in the disk), so relative paths should be always the choice.

ederign:
@jesuino could you please list the env. parameters that you are passing for the editors in order to validate if is what @manstis  and @romartin needs?

jesuino:
Hello @ederign @manstis and @romartin 

Right now we are adding two new parameters: Channel and path.

Since in some channels we may have multiple editors in the same window, the path had to be passed when setting the content, it could not be in a shared object. So we changed the setContent signature to also receive the path as the first parameter.

Channel can be shared across editors, so it is available in a shared `window.envelope.editorContext` object. Only `channel` attribute is available, if you need other information please let me know! Channel can have the values: *VSCODE*, *ONLINE* and *GITHUB*. More will be added later. In this specific PR we are also adding a wrapper object to access editorContext (see EditorContextProvider.java).

Please let me know if you have any question.

manstis:
@jesuino Thank-you for the explanation.

Developers of kogito editors (I think I speak for us all) do not test it in VSCode, Github or elsewhere by default; but prefer to use webapp modules `-webapp-kogito-testing` (launched with `mvn gwt:run`) or `-webapp-kogito-runtime` (WAR built, unzipped and `index.html` simply clicked). These are _standalone_ and do not depend on `kogito-tooling` only `appformer`.

It'd be _nice_ if somehow we could have access to a default `Channel` (provided by `appformer`?) that we could use to determine the host of editor. At the moment we use a singleton CDI object `IsKogito` in `kie-wb-common` that returns true/false depending on whether the editor is being hosted in (any) `kogito-tooling` environment or Business Central.

Therefore, I guess the above being said, it'd be nice to be able to dispense with `IsKogito` and use a Channel provided by `appformer` that defaults to Business Central but could later be overriden by `kogito-tooling` to be `VSCODE`, `ONLINE` or `GITHUB` (i.e. add another value for `BUSINESS-CENTRAL` or whatever).

WDYT? FYI @ederign @porcelli.

porcelli:
+1 for @manstis suggestion

romartin:
+1 also on @manstis 's comment - on the editors side we're agnostic to the channel itself, at least for now, and that's a huge benefit.

manstis:
Thanks for the update @jesuino however instead of replicating `IsKogito` verbatim IMO I'd preferred to simply have had [this](https://github.com/kiegroup/appformer/pull/873/files#diff-fd53705cec4502ab27a6abd5ac35056cR32) return `BUSINESS-CENTRAL` (or whatever suitable constant is best) if the _envelope_ is not available (i.e. not running with a `kogito-tooling` envelope) and remove the new addition of `IsKogito` all together. This would work better when the editors move to whatever the cloud-enabled replacement to Business Central materialises.

We'd need to refactor `kie-wb-common` accordingly to use `EditorContextProvider` instead of our existing `IsKogito` _environment detection mechanism_ (that would be deleted).

@porcelli @ederign @romartin @gitgabrio WDYT?

appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/context/impl/EditorContextProviderImpl.java
```diff
@@ -0,0 +1,48 @@
...
+@ApplicationScoped
+public class EditorContextProviderImpl implements EditorContextProvider {
+
+    @Override
+    public Optional<String> getChannel() {
+        return nullSafe(EditorContextWrapper::getChannel);
+    }
+
+    @Override
+    public boolean isKogito() {
```
gitgabrio: @jesuino 
Hi. I think this should be removed/avoided. 
If "WindowRef.isGwtEditorBeansAvailable()" mean you have the `gwtEditorBeans` variable, this one may be not instantiated for a lot of different reasons - i.e. some error in code implementation/compilation, that are not related to the environment itself.

gitgabrio:
@manstis Thanks for letting me know of that, since also ScenarioSimulation team is working on kogito integration.

@jesuino @porcelli @ederign @romartin @danielezonca @yesamer 
About the "IsKogito" stuff, we already discussed about the removal of that "IsKogito" stuff (it was just a quick workaround), then I agree with @manstis suggestion. 
One little concern I have is that anyway it will create a sort of "matrix". If I did not misunderstood: on one side we have different modules coded/built for specific environments (ie. we have a kogito-testing for development, a kogito-runtime that is actually client-side, and the business-central one) and at the same time those modules are going to ask about the running environment to the container. So - _by design_ - nothing prevent that a kogito-runtime receives a "BUSINESS-CENTRAL" channel, or the business-central module may receive a "VSCODE" channel: it all depends on the way they different pieces have been built and put together.

jesuino:
@manstis @gitgabrio Hello, I will update this commit soon, I was testing alternatives on top of this PR while discussing how to implement how to certify that the editor is running on BC. I will amend this commit for now to avoid further confusion

appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/context/impl/EditorContextProviderImpl.java
jesuino: Thanks for the feedback, @gitgabrio

appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/context/EditorContextProvider.java
```diff
@@ -0,0 +1,50 @@
...
+        return nullSafe(wrapper -> wrapper.getChannel());
+    }
+
+    private Optional<String> nullSafe(Function<EditorContextWrapper, String> action) {
+        EditorContextWrapper wrapper = EditorContextWrapper.get();
+        if (wrapper != null) {
+            return Optional.of(action.apply(wrapper));
+        }
+        return Optional.empty();
+    }
```
danielezonca: I think it is more readable as

```java
 public Optional<String> getChannel() {
      return Optional.ofNullable(EditorContextWrapper.get()).map(EditorContextWrapper::getChannel);
 }
```

gitgabrio: > I think it is more readable as
> 
> ```java
>  public Optional<String> getChannel() {
>       return Optional.ofNullable(EditorContextWrapper.get()).map(EditorContextWrapper::getChannel);
>  }
> ```

@jesuino One question: what does Optional.empty would mean from a functional point of view? Is there a possible situation where the "channel" is empty ?

manstis: Presumably, given the current implementation, the _channel_ would be empty when `IsKogito` is `true`. However I'm proposing `IsKogito` is killed-off and _channel_ **always** returns a value... but let's wait for @jesuino to advise.

jesuino: What I was testing yesterday was something  like:

If you are running Kogito and in a channel, so: isKogito==true and channel==channel where the editor is running
If you are running on Business Central or in a webapp: isKogito==false
When you test the editor in the browser (without any channel), channel.isEmpty() == true

Now adding BUSINESS-CENTRAL channel we simply would have an additional channel for Business Central only or channel.isEmpty ==true if no channel. This is what I am working right now.

manstis: As long as we only have one place to check the environment in which an editor is running I'm happy... i.e. kill off `IsKogito` everywhere. My preference is to simply use `EditorContextProvider.getChannel()` to determine _where_ the editor is running, be it VSCode, Chrome, Business Central or elsewhere.

gitgabrio: @jesuino @manstis 
I think your comments may be merged in one solution (correct me if I'm wrong):
1) always return the channel identifier in one single place
2) do not rely on IsKogito (as you see, using it lead to a matrix of cases)

I personally agree, but still not really like the overall idea that somehow the content (i.e. the editor) should know about the container to work correctly - it seems against the concept of modularity - I have the impression there is some architectural issue behind that this "channel" information is meant to solve.
Beside that, `Optional.empty()` is a nice way to say `null`, and `null` has not any semantical meaning: I would prefer to always return a value or throw an exception.

manstis: My understanding is that _channel_ may be used to determine where to show docks and which may be open by default etc based upon the container/channel... we could achieve this by composing different plugins/extensions for different containers/channels but IMO that could all too easily explode into a maintenance nightmare! There may be other uses of which I am unaware..

gitgabrio: @manstis 
If I follow correctly, that would lead to the following scenario:
1) editor must invoke the "channel" to know where they are running
2) basing on that, the editor itself must implement a sort of "if" - based on the channel, the editor should behave/look differently
3) this could go far away of the simple "screen container" approach I've understood from the beginning - it would require to spread the consequences of such "if" all over the code.
But what I still fail to understand is: we already have two specific modules - one is meant for kogito (and with it I mean client-side only - maybe this is my misunderstanding) and the other for business central. What other differentiation could possible be? *If* we are saying that some behavior of an editor inside VSCode must be different from the github extension, then we are saying that potentially for every new possible channel a different implementation should be provided, and that would lead to nightmare too, IMHO. 
I think that editors should be thought as composable pieces - with the wrapper layer provided only to allow the client-side requirements. Beside that, they should be absolutely agnostic about the container they are running in.

manstis: Hi @gitgabrio IDK the full purpose of _channel_ other than what I've said above that _some_ editors in _some_ channels _may_ need to behave differently - and I don't mean offer different feature sets.. We have 2 wrappers. Plain and simple. One for (current) Business Central and one for kogito (be it VSCode, GitHub etc). The feature sets in each wrapper remain constant however I've heard mention of, for example (and I may be mis-quoting), the Properties Panel (in DMN and BPMN) may be required to be open by default in VSCode but not GitHub.

manstis: If _channel_ is meant to be providing the meta-data for something else perhaps @jesuino or someone in foundation can explain it's exact purpose.

gitgabrio: @manstis @yesamer 
So, if I understand correctly, that _channel_ has some meaning only inside the client-side editors.
Having said that, IMO it looks much more like a "configuration" parameter, that should be set once and for all depending on the environment, and not something to be continuously invoked at runtime.  With that said, I think it would be better to move that during "compilation/build" (or whatever) phase: define a "channel" property that is set when the code is embedded in the container (VSCode, other) and then the editor behave differently based on that variable. I think the behavior would be much easier to predict and the code easier to test (not saying avoiding unpredictable method invocation depending on who-know-what mechanism)

ederign: Hi everyone, 

EditorContextProvider.getChannel()  will return an Enum with

- DEFAULT (on any web app running based on AppFormer);
- On Kogito we will return VSCODE|GITHUB|Online based on the channel that the envelope is running;

Our webapps to ensure the correct values are returned in -testing and -runtime will have to implement an 
@Alternative for EditorContextProvider that will return the desired channel mock.

I also agree with Michael that is a good idea to remove isKogito() API after this PR gets merged.

Thanks everyone for the discussion. I hope that makes sense and provide value for us.

jesuino: updated PR with commit https://github.com/kiegroup/appformer/pull/873/commits/c6760d6b37ac09966d52feb72b92998a944c5652

romartin:
Hey @jesuino @manstis @gitgabrio @ederign 

I'm a bit lost guys on the concrete class names you mention etc, sorry.. In fact it doesn't matter too much, for me, some of the namings, details, etc - I trust on you guys :) 

But IMO we should be agnostic to kogito, vscode, github, whatever... from the editors side. What I think should be the most important is:
* Considering what we now call "kogito" as our standalone editors, it should not even appear the term "kogito" in classes or modules, it's just the concept that our editors are, be default, standalone applications with no context
* The standalone codebase should provide the right interfaces to make features extensible
* Build modules on top of the "standalone" editors for concrete environments, by extending the common/standalone interfaces
* Finally we will be able to provide concrete packages for each environment, and code will be properly isolated between envrionments as well

So IMO is fine the kogito tooling provide some basic APIs we can use for our standalone editors, but rather than this, once providing the ability to get the channel (`EditorContextProvider.getChannel()`), for example, we'll be opening a door for doing tricky stuff and crappy code....

ederign:
Hi @romartin, we designed this with the first use case to be able to decide if we should open or not a dock depending on the channel. If we follow your suggestion, what would be the most suitable model to add this code?

ederign:
Besides that, @romartin  do you agree to split this discussion on two topics?
1-) The plumbing API from foundation (what this PR is about)
2-) How editors will integrate with this API 

For [1] I believe William solution is already done. For [2] we can discuss in another mail thread. 

wdyt?

romartin:
@ederign yeah sure, at some point, decisions depending on the channel has to be done, but I would prefer this decisions being done at kogito tooling level, so on the editors side we should no need to make decisions based on channels. In case of docks, I understand docks are part of the base tooling libraries, so it's tooling who could decide just whether to open or not depending on the channel. I mean, from the editors side, we just managethe content and related events for docks, but I would prefer not making visible the channel to everybody, as people may be tempted to write code for different environments in the same class, so we'll end up with mixing modules and having also code not really specific for the channel in use, but anyway present and doing noise.
I don't want to break your stuff guys, at all, just that giving my expierence (I did lot of crappy code too)  as soon we can identify and provide only the minimum amount of information on our base stuff much better - we want only this information for docks, in case there exist some solution on the tooling side, that makes this information not available yet for others, will reduce the amount of issues and crappy code  that can be used on top. On other thand, I don't really have much detail as you guys on tooling and editors "core" part, and I also understand it's not easy to merge all our acual code to fit new stuff, so I'm just talking at high level and trying to let others think about it when adding new interfaces, because we're a big time and things are growing too fast... you know guys. 

So sorry if you took this as a complain, I don't want to prevent it for being merged, I'm just sharing my high-level vision with you, and tryihng to be all in same page.

Thx guys anyway!!

jesuino:

gitgabrio:
@ederign @manstis @romartin @yesamer 
Hi guys, having read all the previous comments, I have the impression we should take a common decision taking in account all points of view - it also seems to me that somehow the "what is needed" and "how is implemented" get mixed in the discussion.
But I also think it would be better to have a meeting/discussion by voice - once and for all, even if longish - instead of a mailing thread that may last longer then expected and may be confusing to follow. Wdyt?

manstis:
@gitgabrio Personally, I'm happy with this latest proposal. 

The requirement is to have a means to determine the _environment_ in which the editor is running. The implementation here and related PRs does such and allows me to force/override a _channel_ for our `-testing` and `-runtime` webapp.

That said I'm more than happy to attend a meeting with the `kogito-tooling` team and yourself to discuss further before these are merged..

gitgabrio:
@manstis 
It is really possible that everything is in the best shape, but I'm a little bit confused about some details - maybe just because I do not know the internals - that's why I would like to have the chance to discuss them before merging.

appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/context/EditorContextProvider.java
```diff
@@ -0,0 +1,34 @@
...
+ * Provide access to EditorContext properties
+ *
+ */
+public interface EditorContextProvider {
+
+    /**
+     * Access the channel where the editor is running (e.g. ONLINE, GITHUB or VSCODE).
+     * 
+     * @return
+     * The channel where the editor is running or an empty optional if no channel is available.
```
jomarko: `KogitoChannel` is an enum. Not sure `an empty optional` can be returned as we state in javadoc.

appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/context/EditorContextProvider.java
```diff
@@ -0,0 +1,34 @@
...
+     */
+    KogitoChannel getChannel();
```
jomarko: `KogitoChannel` is an enum. Not sure `an empty optional` can be returned as we state in javadoc.

jesuino: fixed

jesuino:
thanks @jomarko I updated the comments

porcelli:
Guys, I think this is going out of proportion :)

the goal of this ensemble is simple: provide a bit of more information, if available, to editors. 

Editors may not even those, but for the practical purpose, they're accessible if needed (and exists).

We should try to keep it as simple as possible, so "optional" parameters are the best way to handle that. Yeah.. java doesn't have the optional parameter concept, but TS/JS world has and this is an integration piece with this world.

@gitgabrio if the concepts are not clear, feel free to contact me to clarify it. But I'm not sure if it's worth a call with all engineers involved to discuss the internals of the implementation. 

@romartin yeah, it may open the door for crappy code full of IFs and ELSEs by channel, so we have to trust the best judgment of the editor's leads.

If the implementation looks buggy or wrong, we should fix it. But I think the review at this point has been a little bit over it.

manstis:
@romartin How the flag is used in editors is up for discussion.. indeed @karreiro and I were just today discussing _how_ we're going to support different model versions in the same editor. DMN has to support 1.2 and 1.3 (soon) that brings new UI requirements.. We could litter our code with `if.. else` or we could have a single factory type singleton that can (CDI) _provide_ different implementations of things for different channels or we could no doubt do something else. @karreiro and I are leaning towards a plugin mechanism where UIs are composed of plugins to provide features (for different model versions). This is comparable to programmatic decisions regarding channels.

Even before this new flag there's nothing stopping somebody else polluting a module with code relating to a different runtime environment (e.g. a Stunner -core module with code expecting to be ran in VSCode etc).. We, as leads, need to remain vigilant in our reviews to police separation.

jomarko:
console output:
```
16:40:13 [INFO] WARNING: You are currently running a version of TypeScript which is not officially supported by typescript-estree.
16:40:13 [INFO] 
16:40:13 [INFO] You may find that it works just fine, or you may not.
16:40:13 [INFO] 
16:40:13 [INFO] SUPPORTED TYPESCRIPT VERSIONS: >=3.2.1 <3.6.0
16:40:13 [INFO] 
16:40:13 [INFO] YOUR TYPESCRIPT VERSION: 3.6.2
16:40:13 [INFO] 
16:40:13 [INFO] Please only submit bug reports when using the officially supported version.
```

appformer-kogito-bridge/src/test/java/org/appformer/kogito/bridge/client/context/KogitoChannelTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class KogitoChannelTest {
+
+    @Test
+    public void withNameTest() {
+        assertEquals(KogitoChannel.GITHUB, KogitoChannel.withName("GitHub"));
```
jomarko: Please cover also `ONLINE`, `VSCODE`, `DEFAULT`

ederign: @jomarko can we fix this in a later JIRA? @jesuino is on PTO and I don't want to block the PR.

ederign: https://issues.redhat.com/browse/KOGITO-866

ederign:

tiagobento:

tiagobento:

PR #875, opened 2020-01-06
[AF-2058] Fixing incomplete logic for setting k8s filesystem provider as default if it is installed

rhtevan:
Signed-off-by: Evan Zhang <evan.zhang@redhat.com>


rhtevan:
@porcelli @adrielparedes @sutaakar Found and fixed an issue of setting k8s filesystem provider as default when it is installed and activated. Could you help review it, thanks!

PR #886, opened 2020-02-03
KOGITO-862: Adding bridge to new resource content get method with Options

jesuino:
Bridge GWT with the new ResourceContentService Options support. 

Right the only supported option is the content type, which can be `text`, the text representation of a content or `binary`, where the content binary is encoded in a base64 string. To retrieve binary content use:

~~~
resourceContentService.get("/path/to/resource", Options.binary());
~~~

Merge with: https://github.com/kiegroup/kogito-tooling/pull/60

### DIFF
--- a/appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/resource/interop/Options.java
+++ b/appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/resource/interop/Options.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.appformer.kogito.bridge.client.resource.interop;
+
+import jsinterop.annotations.JsOverlay;
+import jsinterop.annotations.JsPackage;
+import jsinterop.annotations.JsProperty;
+import jsinterop.annotations.JsType;
+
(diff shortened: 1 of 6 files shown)


jesuino:
@romartin Hello Roger,

We are bringing support for binary content in this PR. The binary content is returned in a base64 string, which can used directly with images.

Would you please kindly let us know our thoughts about this?

Thanks.

jesuino:
@domhanak Would you please review this? Thanks!

jesuino:

appformer-kogito-bridge/src/main/java/org/appformer/kogito/bridge/client/resource/interop/Options.java
```diff
@@ -0,0 +1,63 @@
...
+@JsType(isNative = true, namespace = JsPackage.GLOBAL, name = "Object")
+public class Options {
+
+    /**
+     * The a content type. When "text" the resource content is returned in text format. </br>
+     * When "binary" then the resource content is returned in a base64 encoded format.
+     * @param type
+     */
+    @JsProperty
+    native void setType(String type);
```
romartin: hey @jesuino - not sure if makes sense here, but could the value for the `type` be something more spec-like? I understand calling the resource API, by using whatever options, will finally produce an HTTP request, so maybe can we document properly, in the JavaDocs, and also consider the value for the `type`, for example, as a valid [MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types)?

jesuino: Hello Roger, thanks for the feedback.

The calls to the resourceContent API are not actually HTTP requests. They use the envelope API to exchange messages with the channel. 

A call to get a resource fire an async message to retrieve a certain resource content, captured by the envelope messaging API. Later the channel respond with the resource content, so the API finds the original callback to get the resource and call it with the resource content.

If you want a binary resource, for example, we will encode it using base64 because "Base64 is particularly prevalent on the World Wide Web[1] where its uses include the ability to embed image files or other binary assets inside textual assets such as HTML and CSS files."[1]

Please let me know your thoughts.

Thanks!

[1] https://en.wikipedia.org/wiki/Base64

romartin: ok @jesuino , it makes sense, let's keep it as it is, thx for clarification!

jesuino:

PR #891, opened 2020-02-10
AF-2423: DebugTest.testObjectId rise an error OutOfMemoryError: Java …

nibub17:
AF-2423: DebugTest.testObjectId rise an error OutOfMemoryError: Java heap space

Test intermittently fails in Jenkins (for JVM version of IBM Jdk1.8 ) with java heap space error. 
Whenever System.identityHashCode() returns a negative value, It goes to infinite loop and throws up  "OutOfMemoryError: Java heap space"

### DIFF
--- a/uberfire-api/src/main/java/org/uberfire/debug/Debug.java
+++ b/uberfire-api/src/main/java/org/uberfire/debug/Debug.java
@@ -44,18 +44,19 @@ public class Debug {
      */
     public static String toMemorableString(long num) {
         StringBuilder memorable = new StringBuilder();
+        long hashNum = Math.abs(num);
         for (; ; ) {
-            int chunk = (int) (num & 0xf);
+            int chunk = (int) (hashNum & 0xf);
             memorable.append(CONSONANTS[chunk]);
-            num >>= 4;
-            if (num == 0) {
+            hashNum >>= 4;
+            if (hashNum == 0) {
                 break;
             }
 
-            chunk = (int) (num & 7);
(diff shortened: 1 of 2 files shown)


uberfire-api/src/main/java/org/uberfire/debug/Debug.java
```diff
@@ -44,6 +44,7 @@
      */
     public static String toMemorableString(long num) {
         StringBuilder memorable = new StringBuilder();
+        num=Math.abs(num);
```
jomarko: formatting should be `num = Math.abs(num);`. Even more, from my point of view it would be better not to change value of the method input parameter and just rewrite the line in cycle `int chunk = (int) (Math.abs(num) & 0xf);`
WDYT?

nibub17: Thanks @jomarko , formatting change is done.

Since method uses input parameter as variable and we need to change at multiple lines.
or else we need to assign input parameter to a new variable and use it.
i have added new variable and updated the code. Plz review it

ederign:
Please jenkins execute full downstream build

ederign:
@adrielparedes could you please take a look at this?

adrielparedes:
@nibub17 I've just realized that the commits are not squashed, can you do it so we have one single commit? (if you don't know how to do it ping me back and we can do it together)

adrielparedes:
+1 @ederign this is ready to be merged

PR #893, opened 2020-02-12
DROOLS-3276: [DMN Designer] All GRIDS: Add support for resizing columns using header

manstis:
See https://issues.redhat.com/browse/DROOLS-3276

This PR adds support for resizing columns in the header too (currently you can only resize columns in the body). It also improves use by not selecting the cell next to the pointer when resizing a column (currently after the column resize drag ends a cell is often selected next to the pointer).

### DIFF
--- a/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/NodeMouseEventHandler.java
+++ b/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/NodeMouseEventHandler.java
@@ -20,6 +20,8 @@ import java.util.Optional;
 
 import com.ait.lienzo.client.core.event.AbstractNodeMouseEvent;
 import com.ait.lienzo.client.core.types.Point2D;
+import org.uberfire.ext.wires.core.grids.client.widget.dnd.GridWidgetDnDHandlersState;
+import org.uberfire.ext.wires.core.grids.client.widget.layer.GridLayer;
 
 /**
  * Defines a generic handler for any type of {@link AbstractNodeMouseEvent}.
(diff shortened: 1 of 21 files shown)


uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/NodeMouseEventHandler.java
```diff
@@ -81,4 +83,28 @@ default boolean handleBodyCell(final GridWidget gridWidget,
...
+        final GridLayer gridLayer = (GridLayer) gridWidget.getLayer();
+        final GridWidgetDnDHandlersState.GridWidgetHandlersOperation operation = gridLayer.getGridWidgetHandlersState().getOperation();
+        switch (operation) {
+            case NONE:
+            case COLUMN_RESIZE_PENDING:
+            case COLUMN_MOVE_PENDING:
+            case COLUMN_MOVE_INITIATED:
+            case ROW_MOVE_PENDING:
+            case ROW_MOVE_INITIATED:
+            case GRID_MOVE_PENDING:
```
jomarko: Reading the javadoc. What we mean by Drag and Drop? My understanding was button pressed and moving mouse what would mean here we should have just constants with MOVE in name, not PENDING or INITIATED. So is Drag and Drop also not moving mouse but keeping the mouse button pressed?

jomarko:
### Manual tests
Both on Firefox, Chrome
- [ ] resize columns of DMN decision table
- [ ] resize columns of DMN relation
- [ ] resize columns of Guided Decision table
- [ ] reorder columns of DMN decision table
- [ ] reorder columns of DMN relation
- [ ] reorder columns of DMN context
- [ ] reorder columns of Guided Decision table
- [ ] reorder rows of DMN decision table
- [ ] reorder rows of DMN relation
- [ ] reorder rows of DMN context
- [ ] reorder columns of Guided Decision table
- [ ] SceSim not affected

uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/NodeMouseEventHandler.java
manstis: Hi @jomarko a DnD operation is indeed any mousedown, mousemove, mouseup sequence of events. Furthermore a click event is a combination of mousedown, mouseup on the same HTML element (with or without a mousemove in between). Browsers detect the mousedown, mouseup and raise a separate click event in addition to the mousedown and mouseup events.

`_PENDING` means the pointer is _over_ a header cell (or "resize hot spot" etc) but a mousedown event has not happened (i.e. User is simply moving the pointer over a cell). `_INITIATED` means a previous `_PENDING` operation was detected and now a mousedown event has occurred and the DnD operation has been _initiated_ (but may result in nothing happening, if the User does not actually move the column). When the column has actually been moved the DnD state is moved to `_MOVE`.

Therefore, when single left clicking on a header (to edit it), mousedown, mouseup and click events are raised by the browser. This led to the "edit header" popup showing after _moving_ a column to a new position which was undesirable. To prevent this from happening I added the `COLUMN_MOVE_INITIATED` state that means the cursor was over a movable column and a mousedown event was received BUT the column has NOT been moved to a new position.

In the `isDNDOperationInProgress(..)` [code](https://github.com/kiegroup/appformer/pull/893/files#diff-c4a591df74dd075a2c1ae794a9aa2bb2R102) `COLUMN_MOVE_INITIATED` does not signal a DnD operation has actually happened (i.e. the column has not moved) and therefore both `DefaultGridWidgetCellSelectorMouseEventHandler` or `DefaultGridWidgetEditCellMouseEventHandler` respond to the click event to which they're bound. This is means when a user that a clicks on a header cell it is both selected and its editor is launched.

However IF a column has actually been moved then a DnD operation has actually happened and both of the foregoing handlers do nothing, meaning the column is moved but it is neither selected nor its editor launched.

I hope this explains the change (slightly!)

jomarko:
Done with reviews

manstis:
@jomarko I have responded to your question (presumably blocking approval).

PR #908, opened 2020-02-28
added missing properties

mbiarnes:


mbiarnes:
@manstis appformer seems still to need the strange es_ES.properties

PR #911, opened 2020-03-03
JBPM-9044 - Upgrade kiegroup repos to Wildfly 18.0.1.Final

MarianMacik:


MarianMacik:
jenkins retest this

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-security-management-wildfly/src/main/java/org/uberfire/ext/security/management/wildfly/properties/WildflyUserPropertiesManager.java
```diff
@@ -341,12 +343,13 @@ protected synchronized WildflyGroupPropertiesManager getGroupsPropertiesManager(
         private final PropertiesLineWriterPredicate lineWriterPredicate;
 
         public WildflyUsersPropertiesFileLoader(final String path) {
-            this(path, null);
+            this(null, path, null);
```
mareknovotny: I guess you can't use null here, if you look at that super constructor it doesn't count with it :(
see https://github.com/wildfly/wildfly-core/pull/3784/files#diff-9efd3e0864013b1c72eefb43d5cb8b03R136

mareknovotny:
Btw I have here open kiegroup/appformer/pull/910, so you could play and send PRs to that PR to collaborate ;)

uberfire-extensions/uberfire-security/uberfire-security-management/uberfire-security-management-wildfly/src/main/java/org/uberfire/ext/security/management/wildfly/properties/WildflyUserPropertiesManager.java
MarianMacik: I checked it, you can. Because there is ` if (relativeTo != null) {` before it, which we always have `null`.

PR #927, opened 2020-03-18
DROOLS-5157: [DMN Designer] Monaco editor: Scrolling inside LiteralExpression editor breaks users ability to input data

karreiro:
See https://issues.redhat.com/browse/DROOLS-5157


dupliaka:
jenkins please retest this

karreiro:
Thanks for the review, @dupliaka!

I think we can keep this PR as is for now, and raise another JIRA for scrolling outside the literal expression editor, since this solves the regular/annoying case of scrolling the editor and losing the focus.

When we scroll outside the literal expression editor, we need to move the whole canvas and repaint in the new position, defined by the scroll. Probably this case won't be a trivial fix.

karreiro:
@kiegroup/gatekeepers Can we merge this one?

PR #931, opened 2020-03-20
Added new DESKTOP channel.

paulovmr:


manstis:
@paulovmr @tiagobento @porcelli Can we have DMN's Decision Navigator dock on the left for DESKTOP?!? If you remember it was moved to the right for VSCode so as not to look silly with VSCode's left dock.

paulovmr:
@manstis Yes, it will be on the left, same as the [online editor](dmn.new).

PR #933, opened 2020-03-25
AF-2456: Support Gradual Export in Data Transfer API

jesuino:
This is an extension of Data Transfer functionality to include gradual export.

Now, when clicking on export, a modal will allow users to select what they want to export:

image

When loading files a progress status will be displayed

image

By default all assets will be selected. What will be displayed is:

* Datasets: list of datasets that can be exported;

image

* Pages: List of pages that can be exported;

image

When ok is clicked user is prompted to download export.zip. The export should contains the assets that were selected.

Later, the same asset could be imported and it will override existing content (if conflicting). If the imported content does not exist, it will be created, for example:

* Create two datasets (ds1, ds2);
* Export dataset ds2 only (when exporting,uselect ds1,  unselect pages and set off the navigation);
* Delete ds2;
* Import the exported ZIP;
* ds2 should be back..

The same is also true  for pages and navigation can be restored if it was exported before.


jesuino:
@barboras7 @sthundat Would you please review this PR?

jesuino:

sthundat:
> @barboras7 @sthundat Would you please review this PR?

@jesuino I would review .Thanks

ederign:
@sthundat any news on this PR?

ederign:
@caponetto can you please take a look at this? William can give you a summary.

sthundat:
> @sthundat any news on this PR?

@ederign Currently I am working on AF-2432 .Once the verification done I will take a look on this.
Thanks

dashbuilder/dashbuilder-backend/dashbuilder-services/src/main/java/org/dashbuilder/transfer/DataTransferServicesImpl.java
```diff
@@ -108,22 +114,37 @@ public DataTransferServicesImpl(
...
-    public String doExport() throws java.io.IOException {
-        String zipLocation = new StringBuilder()
-            .append(System.getProperty("java.io.tmpdir"))
-            .append(File.separator)
-            .append(FILE_PATH)
-            .append(File.separator)
-            .append(EXPORT_FILE_NAME)
-            .toString();
+    public String doExport(DataTransferExportModel exportModel) throws java.io.IOException {
+        String zipLocation = new StringBuilder().append(System.getProperty("java.io.tmpdir"))
```
caponetto, line 118: Consider extracting `System.getProperty("java.io.tmpdir")` to a constant. 
I see it being used more than once on this class.

dashbuilder/dashbuilder-backend/dashbuilder-services/src/main/java/org/dashbuilder/transfer/DataTransferServicesImpl.java
```diff
@@ -384,4 +415,70 @@ private File createVersionFile() throws java.io.IOException {
...
+                        files.add(path);
+                    }
+                    return FileVisitResult.CONTINUE;
+
+                } catch (Exception e) {
+                    LOGGER.error(e.getMessage(), e);
+                    return FileVisitResult.TERMINATE;
+                }
+            }
+        });
```
caponetto, line 437: Consider using `DirectoryStream` here.

barboras7:

barboras7:
Hi @jesuino, 

Savitha had a sick day today so I started looking at the PR. Following are my observations/concerns:

1. it is possible to **export** the data even when **nothing is selected** (this is mostly UX issue but IMHO it would be fine if the Ok button would be disabled in such case or some similar approach would be good)

2. the button for enabling/disabling navigation export looks a bit strange; if that's by design from the UX team, you can ignore my comment but to me, it would be more "natural" to **first** see the explanation (**Export navigation**) and on the **right side**, the **ON/OFF** button (also, the "Export navigation" could be in a bigger font to be more visible)
Screenshot from 2020-04-06 21-42-50

3. In the modal dialog, we have **perspectives** but in the export.zip, it is called **pages** (this could confuse the user)

4. an **Export All** option would be nice :slightly_smiling_face: 

5. Would it be possible to have **the modal resizable/expandable**? For the data sets, it would look better if user was able to see all the columns properly
Screenshot from 2020-04-06 21-48-37


Also, I noticed that you have some empty lines just before the end of methods and classes. I don't know if that is intentional and if it is somehow "against" the code conventions you guys use in our code, I am just pointing it out in case it should not be there. :slightly_smiling_face:

barboras7:
... and the rest of the observations I have:

6. It is possible to export navigation without exporting pages. On the other hand, it is okay to export data sets without exporting navigation. Could the user get some warning in case they are exporting pages (without exporting data sets) or navigation (without exporting pages) that they need to make sure that they export all necessary assets?

7. This point is similar to ^ but not so important I think. When the user is importing dashbuilder data, they may not be aware that the navigation they have in the BC will be replaced by the new (imported) one. For the case when the user has their custom navigation in the business central, they might be warned that the navigation will be replaced by the imported one (although I am not sure if this would be too much work to implement, so just a suggestion :slightly_smiling_face: )

Also, please note that it is not possible to verify this issue on Windows as export/import of dashbuilder data does not work on Windows at all. (cc @ederign if we want to check this on Windows, first the above mentioned issue needs to be prioritized and fixed).

jesuino:
Note: remove navigation export. It is used internally when user has nav components in their pages. The navigation should always be exported

dashbuilder/dashbuilder-client/dashbuilder-cms-client/src/main/java/org/dashbuilder/client/cms/screen/transfer/DataTransferExportPopUpView.html
```diff
@@ -0,0 +1,101 @@
...
+						<h4 class="panel-title">
+							<a data-toggle="collapse" data-parent="#dataSelectionAccordion"
+								href="#collapseThree" class="collapsed">Navigation</a>
+						</h4>
+					</div>
+					<div id="collapseThree" class="panel-collapse collapse">
+						<div class="panel-body assets-panel-body">
+							<div class="input-group">
+								<div id="toggleSwitchContainer" class="form-control"
+									aria-describedby="navigationExportAddOn" style="height: auto"></div>
```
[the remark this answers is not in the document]
jesuino: The Toggle was removed.

We decided to remove the navigation selection because users should not opt the navigation out, otherwise components that refers to the navigation may break.

dashbuilder/dashbuilder-backend/dashbuilder-services/src/main/java/org/dashbuilder/transfer/DataTransferServicesImpl.java
jesuino: I tried using it, The code was clearer:
~~~
        final Filter<Path> dirTest = p -> {
            return pathTester.test(p.toString());
        };
        try (DirectoryStream<Path> dirStream = Files.newDirectoryStream(root, dirTest)) {
            dirStream.forEach(files::add);
        } catch (Exception e) {
            throw new IllegalArgumentException("Error walking directory", e);
        }
~~~

but for some reason it was not going through some paths, specifically pages paths (plugin_content) hence it was not correctly helping me to filter the assets I wanted, so I went back to Files.walk

dashbuilder/dashbuilder-backend/dashbuilder-services/src/main/java/org/dashbuilder/transfer/DataTransferServicesImpl.java
```diff
@@ -181,19 +198,39 @@ public FileVisitResult visitFile(Path path, BasicFileAttributes attrs) {
...
+                                                                               .map(this::getFileContent)
+                                                                               .map(this::parseDataSetDefinition)
+                                                                               .collect(Collectors.toList());
+        return new DataTransferAssets(datasetsDefs, pages);
+    }
+
+    @Override
+    public Map<String, String> validateExport(DataTransferExportModel exportsModel) {
+        // TODO Auto-generated method stub
+        return null;
```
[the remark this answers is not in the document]
jesuino: This method was removed, I would try to send this PR implementing two JIRAs, but it would be too much work for a PR review, but I forgot to remove this method.

dashbuilder/dashbuilder-client/dashbuilder-cms-client/src/main/java/org/dashbuilder/client/cms/screen/transfer/DataTransferExportPopUpView.html
```diff
@@ -0,0 +1,101 @@
+<div id="dataTransferExportModalBody" class="container-fluid">
+	<div class="row">
+		<div class="col-md-12">
+			<div id="progressIndicator">
+				<p>
+					<span class="spinner spinner-lg spinner-inline"></span> Loading
+					assets...
```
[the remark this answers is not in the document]
jesuino: Thank you for the suggestion, changed to BusyIndicatorView

jesuino:
@caponetto @barboras7 

Thank you so much for your precious review. I took some time last week and this week to do a better job in the gradual export client side:

* Navigation can't be selected, which means that it will be always exported for use in the other application that will run the export (see https://issues.redhat.com/browse/AF-2439)
* I changed to a wizard, which will have UX work later -> https://issues.redhat.com/browse/AF-2472
* At some point we will have to make a severe validation on the exported assets, this should cover some of the use cases you mentioned. Right now user can export whatever they want at their own risk, but the export all button will be still available;
*  I agree with risks exports. Right now the navigation is the most risk component, because it overrides the existing navigation, but the pages and datasets will add to existing content, not remove what is there. Only conflicting assets will be overridden; 
* One important thing that I noticed on Barbora screenshot is that non public datasets were still visible for users, I removed it from the asset wizard, but kept it in the full export.

Guilherme, thanks for your suggestion regarding how I was handling tables, I created a reusable component "AssetsTableView" which made use to switch presenter and reuse the same view. I also reviewed most of the points you mentioned.

Kindly let me now your thoughts with the new wizard for export. 

Thanks!
export_wizard_final

jesuino:

jesuino:

dashbuilder/dashbuilder-client/dashbuilder-cms-client/src/main/java/org/dashbuilder/client/cms/screen/transfer/DataTransferView.java
```diff
@@ -55,6 +54,8 @@
     private Elemental2DomUtil elem2Dom;
     private Event<NotificationEvent> workbenchNotification;
     private PathUrlFactory pathUrlFactory;
+    
+    @Inject PagesTable pagesTable;
```
caponetto, line 58: Are you using this variable? 
Can it be injected on the constructor just like the others?

dashbuilder/dashbuilder-client/dashbuilder-cms-client/src/main/java/org/dashbuilder/client/cms/screen/transfer/export/wizard/ExportWizard.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    @Override
+    public Widget getPageWidget(int pageNumber) {
+        return wizardPages.get(pageNumber).asWidget();
+    }
+
+    @Override
+    public String getTitle() {
+        return "Export Wizard";
```
caponetto, line 87: Consider using i18n for "Export Wizard".

dashbuilder/dashbuilder-client/dashbuilder-cms-client/src/main/java/org/dashbuilder/client/cms/screen/transfer/DataTransferScreen.java
```diff
@@ -70,49 +81,53 @@ public View part() {
...
-
-                }).doExport();
+        callExportService(DataTransferExportModel.exportAll());
+    }
 
-        } catch (Exception e) {
-            view.exportError(e);
-        }
+    public void doGradualExport() {
+        busyIndicatorView.showBusyIndicator("Loading Export Wizard");
```
caponetto, line 92: Consider using i18n for "Loading Export Wizard".

dashbuilder/dashbuilder-client/dashbuilder-cms-client/src/main/java/org/dashbuilder/client/cms/screen/transfer/DataTransferScreen.java
```diff
@@ -127,4 +142,22 @@ public String getExportFileName() {
     public String getImportFileName() {
         return DataTransferServices.IMPORT_FILE_NAME;
     }
-}
+
+    private void callExportService(DataTransferExportModel dataTransferExportModel) {
+        try {
+            dataTransferServices.call((RemoteCallback<String>) path -> {
```
caponetto, line 148: Could this call take noticeable time? If so, consider using `BusyIndicator` here as well.

jesuino:
Hello @caponetto Thanks for your valuable review. I addressed all your comments in my most recent commit

https://github.com/kiegroup/appformer/pull/933/commits/99820b5d5dc802107e6d814671e6632247ea0cf6

Please let me know if I am missing something.

jesuino:

barboras7:

barboras7:

barboras7:

barboras7:
Hi @jesuino,
great improvement :partying_face: 

While I understand that there are plans to have validation implemented in the future, I have some small suggestions that would not require much time to implement. These are just to make sure that the user is aware of the possible data loss/data corruption:

- user should be aware that if they unselect some page/data set, the exported data may not work well in another business central (this might be just a note in the summary view (it might be there always), warning the user to check what they selected); for example: user selects a page to export but they don't select the data set related to that page
- now, the user has no idea that also navigation is exported -> either they should see this information in the summary view or get a notification when importing the data (also, just a note in the summary might help for now)

Also, to me, it was not obvious what "Export" and "Export wizard" could mean. This might be just a thing of preference, but I would expect something like "Export all" and "Custom export" :slightly_smiling_face:

jesuino:
Hello @barboras7 

Thanks again for your great review. I did your suggestions by adding the feature AF-2458, see https://github.com/kiegroup/appformer/pull/933/commits/fcf9c5b4296277f31107e6642cc5d30b8db6fe1b

The validation requires a certain processing:
* For datasets we need to go down for each component, parse it and check if it has a datasource, then check if the exported model contains that datasource;
* There is a very specific validation which is navigration components, which is go down on navigation components(flat a page to each component like it is did for dataset), check if the component is a nav displayer, if so get the navigation group associated to it, go to the navigation, get the pages that are associated to that group and finally check if the pages are all exported. This may get even more complicated if a group contains other group and so on; For this specific case we will add a friendly message on Runtime, to let users know that a page was not imported. If we validate it correctly the validation may take a few minutes to finish

The validation that is being done is:

* Check if all datasets are exported;
* Check if users is exporting at least one page. << this could be discussed further, because that may be the case where user wants to migrate datasets from an installation to another.

Now, notice that I do filter not public datasets, this is important because Runtime does not care about BC internal datasets, it would only bring confusion. In the other hand, the export all includes everything.

The navigation would also be validated, however, users would get confused by selecting what export in the navigation or not, hence the navigation is always exported and filtered on Runtime.

This will all be documented once Runtime is out.

Kindly let me know your thoughts. 

@caponetto Would you please review the last commit I did? I included validation for the exported assets.

Thanks!

This is how the export looks now:

export_with_validation

jesuino:

barboras7:

jesuino:
thanks, @caponetto

jesuino:

jesuino:
@barboras7 I am in the middle of the build, it should finish in about 30 minutes

jesuino:
hello @barboras7 I uploaded the WAR here, but it seems to be not working in my wildfly 14


https://drive.google.com/file/d/1gikKtTKRAuLTM2vauvtwPK9qarIZe_im/view?usp=sharing

jesuino:

barboras7:
@jesuino thanks William. It’s not working for me neither 😕I was able to build it locally just now as well but no success running it on wildfly... no idea what’s wrong

jesuino:

jesuino:

jesuino:
Hello @barboras7 I finally managed to get a working WAR. I tested on Widlfly 14.0.1 and I also had to increase the heap memory to 2g
You can change the memory settings (Xms and Xmx) on ./bin/standalone.conf or when starting widlfly (didn't test the command below, but it should work)

./bin/standalone.sh -c standalone-full.xml -DXmx2g -DXms2g

Here's the link to download the war, just download it and drop in standalone/deployments directory.

Thanks.

Link to download the WAR https://drive.google.com/file/d/1E90NHWxn_NJy2cotfGO_HafNcWfPDyPU/view?usp=sharing

jesuino:

jesuino:
Hello @barboras7 ,
Thanks again for your feedback.

Would it be possible to share what is in the "Error details"? I tried to reproduce on Dashbuilder but I couldn't

Thanks.

barboras7:
Sure, this is from the server log:
`16:55:05,863 ERROR [org.kie.workbench.common.services.backend.logger.GenericErrorLoggerServiceImpl] (default task-5) Error from user: testadmin Error ID: 377159183 Location: DataTransferPerspective|DataTransferScreen Exception: Uncaught exception: null`

and [here](https://pastebin.com/3DezNq5M) is the error details from UI.

dashbuilder/dashbuilder-backend/dashbuilder-services/src/main/java/org/dashbuilder/transfer/DataTransferServicesImpl.java
```diff
@@ -108,22 +114,37 @@ public DataTransferServicesImpl(
...
+
+        Predicate<Path> readmeFilter = p -> p.toString().toLowerCase().endsWith("readme.md");
+        Predicate<Path> datasetsFilter = def -> true;
+        Predicate<Path> pagesFilter = page -> true;
+        boolean exportNavigation = true;
+
+        if (!exportModel.isExportAll()) {
+            datasetsFilter = filterDatasets(exportModel.getDatasetDefinitions());
+            pagesFilter = filterPages(exportModel.getPages());
+            exportNavigation = exportModel.isExportNavigation();
```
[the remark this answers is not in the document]
jesuino: I would let it in case we want to allow users to select this option in future. For example, I did create [1] so users could export a single page along its dependencies so they can it on runtime. In that case navigation would not be needed.

[1] https://issues.redhat.com/browse/AF-2459

jesuino:
Hello @barboras7 

Thanks again for your feedback. Possibly the issue you faced is related to the one that Eder pointed me yesterday (click next, next, next and a NPE is thrown). In any case I added an extra null checking to the method that was referenced in the stack trace. My answers are inline:

1) I added help to the navigation, thanks for the suggestion!
2) I remapped to the dataset name, thanks!
3) The idea is that these custom exports could be used in Dashbuilder Runtime. In this case, each export should be self contained and contains dashboards, so User should not export only datasets, but also pages, because it is what Dashbuilder Runtime was built for. We can revisit this in future, but for now IMO users should not leave pages behind or it will behave weirdly on runtimes;
4) Thank you, I did improve the messages and made a more specific message for each case.

I also removed the WARNING state from the summary page. Now it has three states: empty, error and success. The result can be seen on the gif below

export_gradual_wizard_improvements


I will trigger a new build, but if it does not work I will build a WAR

jesuino:

barboras7:
Thanks @jesuino! Looks really great :partying_face: 

Anyway, I am still getting the error that I mentioned in my previous comment. I realized now that the error might be related to the fact that I am using some example values provided by the business central (and not my data set). Here's the [link](https://drive.google.com/file/d/1IP143inBxZV1mNazur0W20g3CwCUQe8u/view?usp=sharing) to the video reproducing the issue.
However, I don't think that this is a blocking issue, therefore I can just create a Jira for this if that's the preferred approach so that the PR is not blocked :slightly_smiling_face:

ederign:
@barboras7 let's create a JIRA and handle this in a separate PR. If that is ok to you, can you please create the JIRA and also approve the PR?

jesuino:
Hello @barboras7 ,

Thanks for the video, it clarified the issue. It happens because the sample dataset is never exported, it is an inline dataset, which should not be validated. I added a new commit  (https://github.com/kiegroup/appformer/pull/933/commits/ba0aeddff9cf3e9aa7f70b050c87eaafc23e525a) which avoids null dataset lookups and improve error handling in the wizard.

Thanks!

jesuino:

jesuino:
Hello @barboras7 

There's a WAR built in the CI servers, would you kindly review it? Thanks!

https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/KIE/job/master/job/upstream-downstream/job/appformer.downstream/40/artifact/kiegroup_kie-wb-distributions/business-central-parent/business-central-distribution-wars/business-central/target/business-central-7.38.0-SNAPSHOT-wildfly14.war

jesuino:

jesuino:

jesuino:

jesuino:
The build keeps failing with unrelated error:

https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/KIE/job/master/job/upstream-downstream/job/appformer.downstream/48/console

Let's try one more time, but I would vote for merging if the error is unrelated.

jesuino:

jesuino:

PR #940, opened 2020-04-07
AF-2466: Remove unused dependency of uberfire-nio2-jgit

jomarko:
For more details see https://issues.redhat.com/browse/AF-2466


jomarko:

PR #944, opened 2020-04-17
AF-2467 : Commit history in metrics does not get refreshed when using other than master branch

akumar074:
Ensemble:
[kie-soup#136](https://github.com/kiegroup/kie-soup/pull/136)
[kie-wb-common#3264](https://github.com/kiegroup/kie-wb-common/pull/3264)

### DIFF
--- a/dashbuilder/dashbuilder-backend/dashbuilder-services/src/main/java/org/dashbuilder/dataset/service/DataSetLookupServicesImpl.java
+++ b/dashbuilder/dashbuilder-backend/dashbuilder-services/src/main/java/org/dashbuilder/dataset/service/DataSetLookupServicesImpl.java
@@ -29,6 +29,7 @@ import org.dashbuilder.dataset.DataSetManagerCDI;
 import org.dashbuilder.dataset.DataSetMetadata;
 import org.dashbuilder.dataset.exception.DataSetLookupException;
 import org.dashbuilder.dataset.def.DataSetDef;
+import org.dashbuilder.dataset.uuid.ActiveBranchUUID;
 import org.dashbuilder.dataset.uuid.UUIDGenerator;
 import org.dashbuilder.exception.ExceptionManager;
 import org.jboss.errai.bus.server.annotations.Service;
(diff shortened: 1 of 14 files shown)


akumar074:
Jenkins execute full downstream build please

RishiRajAnand:
@adrielparedes @akumar074  While this implementation associates data sets with current user branch, I suppose there might be several cases where this relation may not be useful.  For example, In space metrics screen,where we show all the metrics irrespective of user branch. IMHO we can also consider maintaining single dataset that contains all the branch metrics and let the clients (UI) filter the records they need. In this way, we can reuse the same data set for space metrics and project metrics (filtered by branch name).

@paulovmr wdyt ^?

akumar074:
@RishiRajAnand I agree with the behavior you suggested for the single dataset. I have a few concerns here. When we create a new branch from a base branch, the new branch has commits of base branch as well. So we get to see a commit message redundantly on the commit history.

akumar074:
Jenkins execute full downstream build please

akumar074:
Jenkins execute full downstream build please

akumar074:
Jenkins execute full downstream build please

akumar074:
Jenkins execute full downstream build please

akumar074:
Jenkins execute full downstream build please

akumar074:
The implementation is changed so will raise a new PR.

akumar074:
> @adrielparedes @akumar074 While this implementation associates data sets with current user branch, I suppose there might be several cases where this relation may not be useful. For example, In space metrics screen,where we show all the metrics irrespective of user branch. IMHO we can also consider maintaining single dataset that contains all the branch metrics and let the clients (UI) filter the records they need. In this way, we can reuse the same data set for space metrics and project metrics (filtered by branch name).
> 
> @paulovmr wdyt ^?

@RishiRajAnand As you pointed out that we should filter the data set on the client-side, I agree to it, It is a much simple solution than what I wanted to achieve here. I have raised a new PR for this.
About the duplication of commits in space metrics, I have filtered the dataset by master for all projects because it seems more appropriate to show only _"stable"_ commits there. 
@paulovmr WDYT^?
Thanks

PR #990, opened 2020-06-18
fixed real directory for Errai Wildfly distribution

mareknovotny:
This is a change for having always directory for Errai Wildfly unpacked distribution to be on the same name as original wildfly distribution as Errai just repackaged it so naming directory to that version is not confusing


mareknovotny:
Jenkins retest this please

PR #1002, opened 2020-07-17
AF-2546: React components for embedded dashboards

jesuino:
This is a requirement for build the react components, it makes the REST API authentication type BASIC instead FORM.

The filter responsible for this was taken from Business Central, so not my code. 

The REST API should be accessible by any role. We can change this later.

To test this change simply access dashbuilder runtime REST API using basic credentials:
```
curl -u 'admin:admin' http://localhost:8080/rest/api
```


jesuino:

PR #1012, opened 2020-07-28
BAPL-678: BPMN Editor - Syntax Highlighting Editor

romartin:
hey @karreiro @LuboTerifaj 

Just a quick improvement to make the Monaco Editor work properly in the BPMN editor.

See https://issues.redhat.com/browse/BAPL-678

Assembly:
- https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1429
- https://github.com/kiegroup/appformer/pull/1012
- https://github.com/kiegroup/kie-wb-common/pull/3372

Thanks!

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client-views-patternfly/pom.xml
+++ b/uberfire-workbench/uberfire-workbench-client-views-patternfly/pom.xml
@@ -536,7 +536,9 @@
                     <include>dev/vs/basic-languages/html/html.contribution.js</include>
                     <include>dev/vs/basic-languages/ini/ini.contribution.js</include>
                     <include>dev/vs/basic-languages/java/java.contribution.js</include>
+                    <include>dev/vs/basic-languages/java/java.js</include>
                     <include>dev/vs/basic-languages/javascript/javascript.contribution.js</include>
+                    <include>dev/vs/basic-languages/javascript/javascript.js</include>
                     <include>dev/vs/basic-languages/kotlin/kotlin.contribution.js</include>
                     <include>dev/vs/basic-languages/less/less.contribution.js</include>
                     <include>dev/vs/basic-languages/lua/lua.contribution.js</include>
@@ -574,13 +576,20 @@
                     <include>dev/vs/basic-languages/tcl/tcl.contribution.js</include>
                     <include>dev/vs/basic-languages/twig/twig.contribution.js</include>
                     <include>dev/vs/basic-languages/typescript/typescript.contribution.js</include>
+                    <include>dev/vs/basic-languages/typescript/typescript.js</include>
                     <include>dev/vs/basic-languages/vb/vb.contribution.js</include>
                     <include>dev/vs/basic-languages/xml/xml.contribution.js</include>
                     <include>dev/vs/basic-languages/yaml/yaml.contribution.js</include>
+                    <include>dev/vs/language/typescript/tsMode.js</include>
+                    <include>dev/vs/language/typescript/tsWorker.js</include>
+                    <include>dev/vs/language/typescript/languageFeatures.js</include>
+                    <include>dev/vs/language/typescript/workerManager.js</include>
+                    <include>dev/vs/language/typescript/lib/typescriptServicesMetadata.js</include>
+                    <include>dev/vs/language/typescript/lib/typescriptServices.js</include>
+                    <include>dev/vs/language/typescript/lib/lib.js</include>
                     <include>dev/vs/editor/editor.main.css</include>
                     <include>dev/vs/editor/editor.main.js</include>
                     <include>dev/vs/editor/editor.main.nls.js</include>
-                    <include>dev/vs/language/typescript/lib/typescriptServicesMetadata.js</include>
                     <include>dev/vs/loader.js</include>
                   </includes>
                 </resource>
--- a/uberfire-workbench/uberfire-workbench-client-views-patternfly/pom.xml
+++ b/uberfire-workbench/uberfire-workbench-client-views-patternfly/pom.xml
@@ -536,9 +536,7 @@
                     <include>dev/vs/basic-languages/html/html.contribution.js</include>
                     <include>dev/vs/basic-languages/ini/ini.contribution.js</include>
                     <include>dev/vs/basic-languages/java/java.contribution.js</include>
-                    <include>dev/vs/basic-languages/java/java.js</include>
                     <include>dev/vs/basic-languages/javascript/javascript.contribution.js</include>
-                    <include>dev/vs/basic-languages/javascript/javascript.js</include>
                     <include>dev/vs/basic-languages/kotlin/kotlin.contribution.js</include>
                     <include>dev/vs/basic-languages/less/less.contribution.js</include>
                     <include>dev/vs/basic-languages/lua/lua.contribution.js</include>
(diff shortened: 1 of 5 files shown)


romartin:
Fixed most of the sonarcloud issues guys, the missing ones are due to native stuff (html, dom, errai, etc). So hope that's everything fine for starting the review again. Thanks!

karreiro:

karreiro:

karreiro:

uberfire-workbench/uberfire-workbench-client-views-patternfly/pom.xml
```diff
@@ -574,13 +576,20 @@
...
+                    <include>dev/vs/basic-languages/typescript/typescript.js</include>
                     <include>dev/vs/basic-languages/vb/vb.contribution.js</include>
                     <include>dev/vs/basic-languages/xml/xml.contribution.js</include>
                     <include>dev/vs/basic-languages/yaml/yaml.contribution.js</include>
+                    <include>dev/vs/language/typescript/tsMode.js</include>
+                    <include>dev/vs/language/typescript/tsWorker.js</include>
+                    <include>dev/vs/language/typescript/languageFeatures.js</include>
+                    <include>dev/vs/language/typescript/workerManager.js</include>
+                    <include>dev/vs/language/typescript/lib/typescriptServicesMetadata.js</include>
+                    <include>dev/vs/language/typescript/lib/typescriptServices.js</include>
```
ederign, line 588: hi @romartin this class has 7.8mb. Is this really necessary?

ederign: Also, it would be nice to review if there is another approach then add all those libraries. 
Unpacked DMN editors before this PR: 33.9Mb
Unpacked DMN editors after this PR: 42.9Mb

romartin: hey @ederign 

Well honestly I cound't make javascript language working property without each one of those added files, it was comtinously complaining and I was adding one by one. On the other hand, it results on typescript also supported, which was not the goal for this task, but the editor doens't work for javascript without those libraries, I tested everything I could... So I think those should be there. yes.

On the other hand:
- We're relying on not minimized packages for the monaco editor, as the production ones are not working properly (thanks to @karreiro for pointing this), so we could try to improve here when some newer library version appears.
- We're adding this libraries for all editors, which is not efficient, agree. 

I can try to play around with POM's and modules and see if I can just include those libraries for the BPMN editor, not at platform level, but honestly don't know yet if it's possible or how tricky it results, will let you know. WDYT? Did you already tested some similar approach?

Thanks guys!

karreiro: Hi @romartin, I understand the issue, and it's not simple indeed :-( As an alternative, we could try to mock the module defined by this file. Wdyt? :-)

karreiro:
Jenkins execute full downstream build.

uberfire-workbench/uberfire-workbench-client-views-patternfly/pom.xml
romartin: what you mean by "mocking" it? :) I'm open for ideas yeah!! :+1:

romartin: As far as I know those javascript modules are not being loaded if not needed, so only BPMN will really load that modules. But it's true that the size for all editors will increase, even if those js modules are not being loaded.

romartin:
Rebased guys

uberfire-workbench/uberfire-workbench-client-views-patternfly/pom.xml
karreiro: @romartin I'm wondering if we could just have something like this (instead of adding `typescriptServices.js`):

```javascript
define("vs/language/typescript/lib/typescriptServices", [], function() { return { /* maybe we'll need a fake API implementation here */ }; });
```

Thus, we would keep Monaco dependency tree happy and avoid this extra 7mb (if this file is not used at runtime).

romartin:
Hey @ederign @karreiro 

First of all thanks guys for your collaboration! 

I'd been thinking on the solution proposed be @karreiro , which is about providing some kinda "mock" files for the specific monaco language libraries. It's a good idea too, but IMO I think this may result on a big cost on mantainance, as we've to provide some "mock" API for each js module, which is not easy and needs for some time, and we're really coupled as well to the code from the third party library, in case they change something maybe we've to also change our implementation or delegation code.

Soo finally, after a few tests, I think I found some intermediate solution which has really low cost from our side - I've reverted the inclusion for specific monaco language modules in the appformer modules, and just inlcude those concrete modules in case of the BPMN editor (see https://github.com/kiegroup/kie-wb-common/pull/3372/commits/fefb00db08f83913fe7bf2128a3909febe013695). This way only the BPMN editor contains those extra libraries, which makes sense as they're required for the authoring. Other editors will remain as they were before, no extra size at all.

I just updated the PRs with those changes, locally it works as exepcted - only BPMN editor has those new js files, other editors are not affected. 

WDYT?

Thanks again for the good brainstorming guys!

romartin:
Again, I'm just proposing some intermediate solution guys, but let's all decide if this one looks enough to go ahead for now. Otherwise do not hestiate to ask for changes at all!! I'm always open to discussion and changes, glad to hear from both guys.

Sooooo Eddeeeer, what you think about last chagnes? I would like to close this during this week guys, if possible, as I'm planning to some PTO....

Anyway thanks in advance for all your collaboration!! <3

CC @ederign @karreiro

ederign:
@romartin sorry for the delay on this. I'm building the online + vscode with our chances here: https://github.com/kiegroup/kogito-tooling/issues/221#issuecomment-673611453

As then we can have more data to choose the best approaches.

romartin:
ok thanks @ederign , plz lemme know asap if some other change should be done here. :+1:

romartin:
rebased from master upstream again

LuboTerifaj (approved):
Hi @romartin 

I didn't find any issues. See my [comment](https://github.com/kiegroup/kie-wb-common/pull/3372#pullrequestreview-476851226) for more information.

Thanks!

romartin:
Hey @ederign @karreiro @porcelli 

I did a quick check to see how much many space the inclusion for the monaco libraries takes.

I locally built the `bpmn-editor` and `dmn-editor` standalone apps (which are used for kogito) for both master and this PR's branch, in order to quickly compare bundle size, and here are the results:

`
8.1M	bpmn-editor-master.war
9.7M	bpmn-editor-pr.war
8.0M	dmn-editor-master.war
8.0M	dmn-editor-pr.war
`
As you can see, DMN editor is not affected by this PR, and the BPMN editor size is increasing 1.6M, it's significant but not much as the numbers that @ederign shared above. Am I missing something @ederign ? 

Anyway, did a quick check for the monaco editor sources, and in general, the minified version is 3 times smaller than the development one, which are using. 

Sooo I assume that doing the work for relying on the `min` bundle may reduce that 1.6M for about 2/3 times, which is interesting... but does it worths then doing the job of fixing the upstream minified version and maintaining that custom webjar? How would it result to assume this size increment until a new offcial version for the library appears?

Thanks!

porcelli:
@romartin for my understanding create a new webjar is a small effort with high benefits. Maybe invest a couple of days (likely that won't need more than that) for such packaging.. so we avoid impact our size (that honestly is already too big for a pure client-side editor).

romartin:
Hey @porcelli 

As commented I'd been doing some quick test on trying to move to the use of minizifed versions for the monaco editor integration, but I've not good feelings tbh:
- By just moving to the use of minified versions, and changing references to those as well, there still appear issues while loading the modules, not trivial to fix. It means we'll have to maybe change our actual code as well, also means we'll have to retest the editors everywhere
- Also as @karreiro pointed out, the minifized version has some bug, so we'll have to fix the bug - not clear even how as the code is being ofuscated, maybe re-minimizing sources (+ any necessary fix) again?

So I propose those actions, in this work, trying to minimize the impact in our bpmn editor:
- Considering the monaco-editor integration "minification" OUT OF SCOPE from this task - keeping this appformer monaco-editor integration as it was before, so we're not worse than before. I see this a new task - a new JIRA for further work on it. Otherwise this task for bpmn may not be ready for the next release...
- As only the BPMN editor is adding some libraries (java,javascript,typescript) - I've properly included the MINIFIED versions for those libraries, instead of the dev ones, so the editor is being increased on the minimum size possible anyway. I've pushed a new commit adding this fix in https://github.com/kiegroup/kie-wb-common/pull/3372

The idea is that the integration itself is not being moved to min versions yet, but at least what we're adding on the bpmn editor side it is, so we're adding as less bytes as possible for this task. This way we're just increasing 800k now, see:

BPMN editor for mater ==> 8.1M
BPMN editor for this PR ==> 8.9M

WDYT??

CC @ederign @karreiro 

Thanks!

ederign (approved):

PR #1044, opened 2020-09-08
AF-2611: Allow default pages in Dashbuilder Runtime

jesuino:
**Thank you for submitting this pull request**

**JIRA**: 

[AF-2611: Allow default pages in Dashbuilder Runtime](https://issues.redhat.com/browse/AF-2611)
[AF-2612: Fix components partition system property](https://issues.redhat.com/browse/AF-2612)

In this PR we address the two JIRAs mentioned above. In AF-2611 we have functional changes, it works

1) A page called "index" will be used when opening a runtime model (import);
2) If you have a single page in your import than this page will be the index;
3) If none of this is done than there's no index/default page.

Thanks!

### DIFF
--- a/dashbuilder/dashbuilder-runtime/src/main/java/org/dashbuilder/client/screens/RuntimeScreen.java
+++ b/dashbuilder/dashbuilder-runtime/src/main/java/org/dashbuilder/client/screens/RuntimeScreen.java
@@ -16,6 +16,8 @@
 
 package org.dashbuilder.client.screens;
 
+import java.util.List;
+
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
@@ -26,7 +28,9 @@ import org.dashbuilder.shared.model.RuntimeModel;
 import org.uberfire.client.annotations.WorkbenchPartTitle;
 import org.uberfire.client.annotations.WorkbenchPartView;
 import org.uberfire.client.annotations.WorkbenchScreen;
+import org.uberfire.client.mvp.PlaceManager;
 import org.uberfire.client.mvp.UberElemental;
+import org.uberfire.ext.layout.editor.api.editor.LayoutTemplate;
 import org.uberfire.workbench.model.menu.Menus;
 
 /**
@@ -38,6 +42,8 @@ import org.uberfire.workbench.model.menu.Menus;
 public class RuntimeScreen {
 
     public static final String ID = "RuntimeScreen";
+
+    public static final String INDEX_PAGE_NAME = "index";
 
     private static AppConstants i18n = AppConstants.INSTANCE;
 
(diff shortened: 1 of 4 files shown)


jesuino:
@caponetto @barboras7 these are the last two changes planned for Runtime before the next release. Would you please review it?

Thanks!

jesuino:
jenkins execute cdb

jesuino:
jenkins execute fdb

dashbuilder/dashbuilder-runtime/src/main/java/org/dashbuilder/client/screens/RuntimeScreen.java
```diff
@@ -67,6 +76,21 @@ public void loadDashboards(RuntimeModel runtimeModel) {
         NavTree navTree = runtimeModel.getNavTree();
         Menus menus = menusHelper.buildMenusFromNavTree(navTree).build();
         view.addMenus(menus);
+
+        goToIndex(runtimeModel.getLayoutTemplates());
```
caponetto, line 80: Can we move this call to the `RouterScreen.java`?
It looks like `loadDashboards` has more responsibilities than it implies.
So it makes more sense to me to let `RouterScreen` call `loadDashboards` and then `goToIndex`.

caponetto (commented):
LGTM, @jesuino. I left only one suggestion.

dashbuilder/dashbuilder-runtime/src/main/java/org/dashbuilder/client/screens/RuntimeScreen.java
jesuino: done, thanks @caponetto

caponetto (approved):

jesuino:
jenkins execute fdb

barboras7 (approved):
The PR looks good, everything's working. 

Just one comment - from a user perspective, it was a bit surprising that if there's some default page (index or a single page in import), there's no more welcome page/home page as the default page replaces it. I would expect that after clicking on the home page button, I would be redirected to the welcome page (but that's just my point of view :) ). If course, I understand this if the default page is considered a home page.

jesuino:
Hello @barboras7 

Thanks again for your review!

The home page is just a place holder because we didn't know where to take users when they didn't select a dashboard. The idea is always have a dashboard loaded! Hence we didn't show it when users know where to take us when clicking on home button.

Also the home is the home for a dashboard (import=xxxx) we also expect users to usually explore a specific dashboard instead going to multi dashboard page. When we have security discussion finished we will probably open a dashboard directly and skip multi selection in cases where user has access to a single dashboard import!

barboras7:
Hi @jesuino,

thanks for the clarification :slightly_smiling_face:

PR #1061, opened 2020-10-09
AF-2636: Dashbuilder Pluggable Service Interface

jesuino:
[AF-2636](https://issues.redhat.com/browse/AF-2636)

This is a major change for components protocol to allow us to plug reusable functions that can be called by components. To achieve this we had first to formalize the messages exchange between custom components and DB, so we introduced Message type and class ExternalComponentDispatcher, that listen for all messages, look for the component and deliver the message.

This is a breaking change because now we support messages of type "INIT", which are sent without DATASET so users can find its own ID.

If a message is a function call then we have a function lookup and its execution. The function is currently made internally, but we could extend it to users so they can register their own function.

To a component invoke a function it must:

* First of all during INIT function retrieve its ID:

```
  window.addEventListener("message", receiveMessage, false);

  function receiveMessage(event) {
    const message = event.data;
    const type = message.type;

    if (type === "INIT") {
      id = message.properties.get("component_id");
    }
  }
```

* Send a function of type `FUNCTION_CALL` with a parameter named `functionCallRequest` and its ID. The parameter functionCallRequest contains all information to invoke a function, including the function name and function parameters:

```
  function callFunction() {
    const props = new Map();
    props.set("component_id", id);
    props.set("functionCallRequest", {
      functionName: document.getElementById("functionName").value,
      parameters: new Map(),
    });
    window.parent.postMessage({
      type: "FUNCTION_CALL",
      properties: props,
    });
  }
```

* Then the component waits Dashbuilder to respond with a message of type `FUNCTION_RESPONSE`. The response will bring a FunctionResponse object with the execution result, a message and the result type, which I will talk more later. The function response also includes the function request object, which can help the component to find which function was invoked if multiple invocations are made:

```
  function receiveMessage(event) {
    const message = event.data;
    const type = message.type;

    if (type === "INIT") {
      id = message.properties.get("component_id");
      console.log("INIT Received. Component ID: " + id);
    }

    if (type === "FUNCTION_RESPONSE") {
      const response = message.properties.get("functionResponse");

      document.getElementById("result").innerHTML = `
            <b>Type:</b> ${response.resultType} <br />
            <b>Message:</b> ${response.message} <br />
            <b>Result:</b> ${response.result} <br />
        `;
    }
    console.log(message);
  }
```

A function call may not be sucessful, in these cases DB will respond either with FUNCTION_ERROR, used when an exception is thrown by the function during execution, or FUNCTION_NOT_FOUND, used when the function name is not found. The message in function response will give more details about what went wrong.

You can see this in action in the following GIF:

function_call_demo

Another message added is FILTER. Now components can request filters. To do this it must send a message of type FILTER carrying a property name "filter" that contains the FilterRequest object:

```
function sendMessage(props) {
    if (!props) {
        props = new Map();
    }
    props.set("component_id", id);
    window.parent.postMessage({
        type: "FILTER",
        properties: props
    });
}
function filter(reset, col, row) {
    var props = new Map();

    props.set("filter", {
        reset: reset,
        column: col,
        row: row

    })
    sendMessage(props);
}
```
See it in action:

filter_request

Important to bear in mind:

* If a message does not contain a component_id, then it is ignored.
* The component is responsible to handle user communication when a function is in execution;
* INIT messages should be responded with a READY, and DB will show a loading state at the component, but ATM we won't be using this or all components would be broken;
* Different from FUNCTION_CALL, FILTER does not respond to the component is something is wrong. It just ignore the FILTER if a request

Finally I also added configuration_issue/configuration_ok, but it is not supported at the moment, we will start using it in some experimental components, but it is not part of the supported API. The way it works is that components can tell dashbuilder if dataset columns are not the expected format and dashbuilder will handle the UI, when it is corrected the component can send a message back to dashbuilder saying that the configuration is okay now, see:

configuration_issue_new

The code for these components can be found in my github:

Call Function: https://github.com/jesuino/dashbuilder-components/tree/master/call_function
Filter: https://github.com/jesuino/dashbuilder-components/tree/master/simplest_component
Bad Configuration https://github.com/jesuino/dashbuilder-components/tree/master/configuration_issue

### DIFF
--- a/dashbuilder/dashbuilder-client/dashbuilder-displayer-client/src/main/java/org/dashbuilder/displayer/client/component/ExternalComponentDispatcher.java
+++ b/dashbuilder/dashbuilder-client/dashbuilder-displayer-client/src/main/java/org/dashbuilder/displayer/client/component/ExternalComponentDispatcher.java
@@ -1,0 +1,179 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.dashbuilder.displayer.client.component;
+
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Consumer;
+
+import javax.annotation.PostConstruct;
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+
+import elemental2.dom.DomGlobal;
+import elemental2.dom.MessageEvent;
+import jsinterop.base.Js;
+import org.dashbuilder.displayer.client.component.function.ComponentFunctionLocator;
+import org.dashbuilder.displayer.client.component.function.ExternalComponentFunction;
+import org.dashbuilder.displayer.client.resources.i18n.CommonConstants;
+import org.dashbuilder.displayer.external.ExternalComponentMessage;
+import org.dashbuilder.displayer.external.ExternalComponentMessageHelper;
+import org.dashbuilder.displayer.external.ExternalComponentMessageType;
+import org.dashbuilder.displayer.external.FunctionCallRequest;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Listen to all component messages and find the destination.
+ *
+ */
+@ApplicationScoped
+public class ExternalComponentDispatcher {
+
+    static final CommonConstants i18n = CommonConstants.INSTANCE;
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ExternalComponentDispatcher.class);
+
+    @Inject
+    ExternalComponentMessageHelper messageHelper;
+
+    @Inject
+    ComponentFunctionLocator functionLocator;
+
+    Set<ExternalComponentListener> listeners;
+
+    @PostConstruct
+    public void setup() {
+        listeners = new HashSet<>();
+        startListening();
+    }
+
+    public void register(ExternalComponentListener listener) {
+        // make the component listener ready by default
+        listener.makeReady();
+        listeners.add(listener);
+    }
+
+    public void unregister(ExternalComponentListener listener) {
+        listeners.remove(listener);
+    }
+
+    public void onMessage(ExternalComponentMessage message) {
+
+        ExternalComponentMessageType type = messageHelper.messageType(message);
+
+        switch (type) {
+            case FILTER:
+                handleFilter(message);
+                break;
+
+            case FUNCTION_CALL:
+                handleFunction(message);
+                break;
+
+            case READY:
+                handleReady(message);
+                break;
+
+            case FIX_CONFIGURATION:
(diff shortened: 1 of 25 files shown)


dashbuilder/dashbuilder-shared/dashbuilder-displayer-api/src/test/java/org/dashbuilder/displayer/external/ExternalComponentMessageHelperTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Mock
+    JsMap<String, Object> properties;
+
+    @InjectMocks
+    ExternalComponentMessage message;
+
+    private ExternalComponentMessageHelper helper;
+
+    @Before
```
caponetto, line 50: Can we use `BeforeAll` instead?

caponetto:
Jenkins run fdb

jesuino:
Hello @caponetto again thanks for the great review. I also included a null checking for message type, without it the unknown messages from components throw a really unreadable exception.

I made the changes in my latest commit.

caponetto (approved):

dashbuilder/dashbuilder-client/dashbuilder-displayer-client/src/main/java/org/dashbuilder/displayer/client/component/ExternalComponentDispatcher.java
```diff
@@ -0,0 +1,179 @@
...
+
+    }
+
+    private void handleOkConfiguration(ExternalComponentMessage message) {
+        findDestination(message, ExternalComponentListener::configurationOk);
+    }
+
+    private void handleConfiguration(ExternalComponentMessage message) {
+        findDestination(message, destination -> {
+            Optional<String> configurationIssue = messageHelper.getConfigurationIssue(message);
```
sthundat, line 113: I would prefer to use the capability of optional instead of specific if else statement.
We could reduce this in a single line of code:
messageHelper.getConfigurationIssue(message)
.ifPresent(x-> destination.onConfigurationIssue(configurationIssue.get()))
orElse(x-> destination.onConfigurationIssue(i18n.componentConfigDefaultMessage());

sthundat: same implementation could apply in multiple places

sthundat:
@jesuino ,
Can you explain how to plug JS functions for the component
Regards,
Savitha

dashbuilder/dashbuilder-client/dashbuilder-displayer-client/src/main/java/org/dashbuilder/displayer/client/component/ExternalComponentDispatcher.java
jesuino: Hello Savitha,

Thanks for the feedback. I modified this part to be an one liner. However, in other parts it does not apply because when the result is not present I have to do operations with other object.

When we support Java 11 I wish I could refactor to use ifPresentOrElse!

jesuino:
Hello @sthundat 

To plug functions we need to write them in GWT/Errai currently. We would make a native register, so users can use Javascript to register their functions, however, we needed to define how it would work and it would delay this feature. 

This specific feature is very important so we can progress to AF-2637 and implement functions that will be required for heatmap so it can retrieve information to build the processes diagram.

For AF-2637 we may have to refactor some parts of this PR and if there's time I will check if I can add native Javascript function registration (like it is done with Kogito currently)

Thanks!

sthundat:
Hi @jesuino ,
Thanks for the clarification.PR looks good .
Regards,
Savitha

jesuino:
jenkins execute cdb

jesuino:
jenkins execute fdb

sthundat (approved):

jesuino:
Test failure not related:

org.kie.hacep.LocalStorageStreamingKieSessionTest.getCommandsTest

PR #1084, opened 2020-12-08
DROOLS-5407 : User can not select cell of Guided Decision Table

Rikkola:
**JIRA**:  

[link](https://issues.redhat.com/browse/DROOLS-5407)

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* <b>a pull request</b> please add comment: <b>Jenkins retest this</b>
 
* <b>a full downstream build</b> please add comment: <b>Jenkins run fdb</b>
  
* <b>a compile downstream build</b> please  add comment: <b>Jenkins run cdb</b>

* <b>a full production downstream build</b> please add comment: <b>Jenkins execute product fdb</b>

* <b>an upstream build</b> please add comment: <b>Jenkins run upstream</b>
</details>

### DIFF
--- a/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseMoveHandler.java
+++ b/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseMoveHandler.java
@@ -159,11 +159,9 @@ public class GridWidgetDnDMouseMoveHandler implements NodeMouseMoveHandler {
                                 renderingInformation,
                                 cx);
 
-            //If over Grid but no operation has been identified default to Grid dragging; when unpinned
             if (!layer.isGridPinned()) {
                 if (state.getActiveGridWidget() == null) {
                     state.setActiveGridWidget(gridWidget);
-                    state.setOperation(GridWidgetDnDHandlersState.GridWidgetHandlersOperation.GRID_MOVE_PENDING);
                 }
             }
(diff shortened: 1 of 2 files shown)


Rikkola:
@jomarko @manstis Since the default is `GRID_MOVE_PENDING` the `if (isDNDOperationInProgress(gridWidget)) {` checks in the Handlers never pass, blocking them from being used. The question now is if the old way or this PR is correct. I don't think this cancels the goal set in the DROOLS-3276.

The old PR that changed the behaviour: https://github.com/kiegroup/appformer/pull/893

jomarko (approved):
Thank you @Rikkola

Rikkola:
@manstis I could use your wise words with this PR :)

manstis:
Hi @Rikkola my fear is that this change means the grids will no longer be draggable in the "graph" version of the editor.

This [condition](https://github.com/kiegroup/appformer/blob/f406a85e800587bf19792fda155da6f8f7565c60/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/dnd/GridWidgetDnDMouseDownHandler.java#L92) will never be met.

Furthermore the `isDNDOperationInProgress(..)` method should be [returning](https://github.com/kiegroup/appformer/blob/f406a85e800587bf19792fda155da6f8f7565c60/uberfire-extensions/uberfire-wires/uberfire-wires-core/uberfire-wires-core-grids/src/main/java/org/uberfire/ext/wires/core/grids/client/widget/grid/NodeMouseEventHandler.java#L105) `false` for `GRID_MOVE_PENDING` and hence the click handler (to select a cell) work OK.

I can dig more if you like, but I'll need to setup my environment.

Happy to help.

Rikkola:
@manstis So maybe this is not the fix. We could also pin by default when just editing one dtable.

PR #1090, opened 2021-01-11
KOGITO-3909: OuiAttributes for Editor Tab Entry

jomarko:
For KOGITO-3909 easier testing this PR add some OuiaAttributes into the Editor Tab entry
For more details see https://issues.redhat.com/browse/KOGITO-3909

Ensemble together with:
- https://github.com/kiegroup/kie-wb-common/pull/3537

Prerequisite for:
- https://github.com/kiegroup/kogito-tooling/pull/372

### DIFF
--- a/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/tab/TabPanelEntry.java
+++ b/uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/tab/TabPanelEntry.java
@@ -173,7 +173,7 @@ public class TabPanelEntry implements HasActive {
 
         @Override
         public OuiaComponentIdAttribute ouiaComponentId() {
-            return new OuiaComponentIdAttribute("editor-nav-tab-" + getText());
+            return new OuiaComponentIdAttribute(getText());
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


karreiro (approved):
Thank you, @jomarko.

uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/tab/TabPanelEntry.java
```diff
@@ -157,5 +165,21 @@ public HandlerRegistration addMouseDownHandler(MouseDownHandler handler) {
...
         }
+
+        @Override
+        public OuiaComponentTypeAttribute ouiaComponentType() {
+            return new OuiaComponentTypeAttribute("editor-nav-tab");
+        }
+
+        @Override
+        public OuiaComponentIdAttribute ouiaComponentId() {
+            return new OuiaComponentIdAttribute("editor-nav-tab-" + getText());
```
jstastny-cz, line 176: Just FYI:
We don't need that prefix, because id is supposed to just distinguish between components of same component-type and on the same level only - it's not supposed to be globally unique across various component types. In other words, it's perfectly fine to have multiple elements same value for data-ouia-component-id if they are either:
* on different levels of DOM tree
* of different data-ouia-component-type

jstastny-cz (approved):
Just a harmless semantic (not-even-)issue. Looks good, we can go ahead with merge if you prefer the current solution.

uberfire-workbench/uberfire-workbench-client-views-patternfly/src/main/java/org/uberfire/client/views/pfly/tab/TabPanelEntry.java
jomarko: Will simplify the id, thank you

jstastny-cz (approved):
Thanks, approval still holds :-)

jomarko:
Jenkins please retest this

PR #1190, opened 2021-08-20
RHPAM-3845 : Imported projects not visible after migration from RHPAM 6.4 to 7.11

RishiRajAnand:
**Thank you for submitting this pull request**

**RHPAM-3462**: _https://issues.redhat.com/browse/RHPAM-3462_ 

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* <b>a pull request</b> please add comment: <b>Jenkins retest this</b>
 
* <b>a full downstream build</b> please add comment: <b>Jenkins run fdb</b>
  
* <b>a compile downstream build</b> please  add comment: <b>Jenkins run cdb</b>

* <b>a full production downstream build</b> please add comment: <b>Jenkins execute product fdb</b>

* <b>an upstream build</b> please add comment: <b>Jenkins run upstream</b>
</details>


uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/config/ConfigurationServiceImpl.java
```diff
@@ -288,24 +287,27 @@ public void endBatch() {
...
-                                                                                                     type.getExt());
+                type.getExt());
 
         //Only load and cache if a file was found!
         final Iterator<Path> it = foundConfigs.iterator();
         if (it.hasNext()) {
             while (it.hasNext()) {
                 final String content = ioService.readAllString(it.next());
                 final ConfigGroup configGroup = marshaller.unmarshall(content);
+                if (namespace != null) {
```
adrielparedes, line 308: In this case, what happens is there is no namespace (or it is null)?

adrielparedes (changes requested):

uberfire-structure/uberfire-structure-backend/src/main/java/org/guvnor/structure/backend/config/ConfigurationServiceImpl.java
RishiRajAnand: @adrielparedes ideally it should fetch the namespace from the repository config file, as a result of unmarshalling on line 307. But we found that in some versions, config file didn't contain the namespace. so if the namespace is null on line 308, this means either the namespace is fetched from the file or the resource doesn't require a namespace at all (like spaces).

RishiRajAnand:
Jenkins run fdb

tomasdavidorg (approved):
Built and tested locally. Looks good to me. 🙂

adrielparedes:
@RishiRajAnand is there any problem with tests?

RishiRajAnand:
> @RishiRajAnand is there any problem with tests?

@adrielparedes the error is org.uberfire.ext.metadata.backend.infinispan.provider.InfinispanPingServiceTest.testPingSuccess

this error can be seen in multiple unrelated PR's.

